PR #5, opened 2015-05-29
Feature: Get video object by plex "key" identifier

simonwjackson:
I'm not sure if library.getByKey() is the best naming convention, but it works :)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,9 +2,13 @@ syntax: glob
 *.db
 *.log
 *.pyc
+*.swp
 *.sublime-*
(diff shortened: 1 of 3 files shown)


mjs7231:
Looks good, thanks! :+1:

mjs7231:
Just a heads up.. In PlexAPI version 3, this function will be moving to `plex.fetchItem(key)` where key is not just the number, but the full relative path (ie: /library/metadata/12345).

PlexAPI v3 is several weeks away.

PR #8, opened 2015-06-01
Feature: Get the URL to stream a video

simonwjackson:
I'm an amateur when it comes to OO design, so i wasnt sure if this was the best way to structure this. If you have any advice please let me know.

```python
jurassic_park = plex.library.section('Movies').get('Jurassic Park')
url = StreamUrl(server=plex, key=jurassic_park.key, max_bitrate=128,
                         max_resolution="1280x720", offset=0)
print(url)

# http://123.456.789.864:10041/video/:/transcode/universal/start?copyts=1&offset=0&path=http%3A%2F%2F127.0.0.1%3A32400%2Flibrary%2Fmetadata%2F8&mediaIndex=0&X-Plex-Platform=Chrome
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,6 +79,7 @@ available libraries.
     jurassic_park = plex.library.section('Movies').get('Jurassic Park')
     url = StreamUrl(server=plex, key=jurassic_park.key, max_bitrate=128,
                     max_resolution="1280x720", offset=0)
+    print(url)
 #### FAQs ####
 
 **Q. Why are you using camelCase and not following PEP8 guidelines?**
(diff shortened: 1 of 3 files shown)


mjs7231:
We should definetly add this.  However, I still deciding if this should in fact be a separate class as opposed to directly ingrained in the Video object.  It  seems quite complex, give me a few days to ponder this. :)

simonwjackson:
I think it should be in the video object as well, i just wanted to get it working first without thinking about any abstraction or inheritance.

munnerz:
I feel `jurrasic_park.stream_url` should return the URL to stream - there's no need to create an object, which in this instance is effectively a (not really) subclass of a String.

mjs7231:
@munnerz That's what I was thinking as well. I have the code (a very simplified version of simonwjacksons version) almost ready to go, but can't actually get it to stream in VLC yet, so need to figure out what is wrong there.

simonwjackson:
@mjs7231 You may be missing a parameter in the URL that is preventing plex to encode the stream.

mjs7231:
I added this without using a separate StreamURL object.  The function Video.getStreamURL was added using your code as the example, thanks!

PR #23, opened 2015-09-05
Python3 fixes / urllib, printing, dict access

trtd:
Python 3 fixes (compatible with Python 2):
* use urllib.parse module for quote() and urlencode() instead of urllib
* printing using parenthesis
* dict access: items() instead of iteritems(), keys() converted to list
* filter(): convert results to list

### DIFF
--- a/examples/examples.py
+++ b/examples/examples.py
@@ -31,7 +31,7 @@ def example_003_list_all_clients(plex):
     for client in plex.clients():
         print(client.name)
     else:
-        print 'No clients'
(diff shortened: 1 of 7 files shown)


adrienbrault:
In order to use this library in https://github.com/balloob/home-assistant , we'd at least need this patch merged. @mjs7231

mjs7231:
@adrienbrault All set, and the few tests I have are all passing. New version 1.1.0 is up on PyPi.

PR #24, opened 2015-09-12
Created query function for client

miniconfig:
Created a query function for the client, and changed the timeline to
use that instead of the server’s.

### DIFF
--- a/plexapi/client.py
+++ b/plexapi/client.py
@@ -4,10 +4,12 @@ See: https://code.google.com/p/plex-api/w/list
 """
 import requests
 from requests.status_codes import _codes as codes
+from plexapi import log
 from plexapi import TIMEOUT, log, utils, BASE_HEADERS
-from plexapi.exceptions import BadRequest
+from plexapi.exceptions import BadRequest, NotFound
 from xml.etree import ElementTree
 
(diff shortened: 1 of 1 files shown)


miniconfig:
I'm not entirely certain I've done this properly (I'm particularly unsure about TOTAL_QUERIES), but I thought I'd give it a shot since I'm the one that needs it.  It does seem to work this way.

mjs7231:
Can you explain what was broken before?  I am a bit unfamiliar with timeline in general.

miniconfig:
It was to fix what I was seeing in my comment [here](https://github.com/mjs7231/python-plexapi/commit/b2c9a29d1e0d9c0242a01f2e574292b7591dea8d#commitcomment-13013198) - Timeline seems to be a function of the client, not the server - before your fix to handle timeout in the connection, it was using the proper URL, but your fix had it (at least with my setup) trying to connect to the server with a URL that exists only in the client interface.

mjs7231:
I did a lot of cleanup in this area in the last few days.  The changes this merge request was solving should also be working. Specifically, I changes the default client actions to talk directly to the client and not proxy through the server. A test for timeline has also been added to ensure this wont regress.

PR #27, opened 2015-10-12
Get devices and implement play/navigation control

adrienbrault:


mjs7231:
I'm still planning to review and pull this, just have been busy these days at home.  Thanks for the patch and sorry for the delay.

mjs7231:
Merging.  I think I can cleanup the duplicate code with Play / Navigation control a bit but I can go that on the main branch. Thanks for the contribution!

plexapi/myplex.py
```diff
@@ -178,3 +181,95 @@ def _findResource(resources, search, port=32400):
...
+        headers = plexapi.BASE_HEADERS
+        headers['X-Plex-Target-Client-Identifier'] = self.clientIdentifier
+        response = requests.get(url, headers=headers, timeout=TIMEOUT)
+        if response.status_code != requests.codes.ok:
+            codename = codes.get(response.status_code)[0]
+            raise BadRequest('(%s) %s' % (response.status_code, codename))
+        data = response.text.encode('utf8')
+
+        if data:
+            try:
```
mjs7231: What's the reason this is in a try/except block?  We're you getting bad XML back form some of the devices?

PR #28, opened 2015-10-28
Plex needs the refresh to be a PUT request

blha303:
Otherwise you get `plexapi.exceptions.BadRequest: (405) method_not_allowed`

### DIFF
--- a/plexapi/video.py
+++ b/plexapi/video.py
@@ -2,6 +2,7 @@
 PlexVideo
 """
 import re
+from requests import put
 from plexapi.client import Client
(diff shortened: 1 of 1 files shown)


mjs7231:
Thanks.  Must have changed in a recent media server build. :+1:

blha303:
Good thing you pulled it after I actually checked it for bugs and squashed the commits ^_^

mjs7231:
Lol I must have loaded the Files Changed tab right after you removed it.  Because whats in there now is what I saw when I merged it.

blha303:
I had `import requests.put` before, cos that's what I had in my project here and I hadn't tested it to see that it obviously wouldn't work.

PR #29, opened 2015-10-31
Add method Episode.is_watched()

sprt:


mjs7231:
Not accepting this one as I did it slightly differently, but the functionality is there now. It's a property rather than a property-method on both Movie and Episode in change 430fb256. Thanks for the initial version! :+1:

PR #34, opened 2015-11-22
Update video.py: self.key already contains slash

bramwalet:
self.key already contains a leading slash therefore this call to the Plex server will fail with 404 not found.
When the slash from the query location is removed, the call will succeed.


mjs7231:
Thanks for the fix, sorry about the bug.

PR #38, opened 2016-01-05
Level for a media object might not always be an integer

logaritmisk:
I have a problem with one episode in my library. For some reason level on the media object isn't a integer (it is a string, 'medium') so I got an exception.

I checked the web interface and they printed it as 'NaN'. So I did the same thing.

I don't know if this is the correct place to add this, maybe this should be in the cast function? So it affect all integers.


mjs7231:
As suggested I implemented this same fix in the cast() function. Thanks for the fix.

commit 8c82651ce1949e849e06bbbe728f5a9a0ccf56b9
Author: Michael Shepanski <mjs7231@gmail.com>
Date:   Mon Mar 14 22:28:21 2016 -0400
Level for a media object might not always be an integer

PR #44, opened 2016-02-14
Add groundwork to have queries like: "actor name and unwatched movie"

danielBreitlauch:
This adds groundwork to have queries like: "actor name and unwatched movie".
It adds MovieSection methods to get Actor/Writer/Genre.. objects from their names.
These can then be used in Section.search() as query arguments.

Changes: 
- Additional MovieSection methods to query by genre and writer (plex directly supports it).
- VideoTag objects can be created from dict.

### DIFF
--- a/plexapi/client.py
+++ b/plexapi/client.py
@@ -3,6 +3,8 @@ PlexAPI Client
 See: https://code.google.com/p/plex-api/w/list
 """
 import requests
+import json
+import urllib
 from requests.status_codes import _codes as codes
 from plexapi import TIMEOUT, log, utils, BASE_HEADERS
 from plexapi.exceptions import BadRequest
@@ -31,6 +33,9 @@ class Client(object):
         self.protocolCapabilities = data.attrib.get('protocolCapabilities', '').split(',')
         self.state = data.attrib.get('state')
         self._sendCommandsTo = SERVER
+        self.id = 1
+        self.rpc_version = "2.0"
+        self.headers = {'content-type': 'application/json'}
 
     def sendCommandsTo(self, value):
         self._sendCommandsTo = value
@@ -40,6 +45,15 @@ class Client(object):
         if sendTo == CLIENT:
             return self.sendClientCommand(command, args)
         return self.sendServerCommand(command, args)
+
+    def sendRpcClientCommand(self, method, params=None):
+        payload = {"jsonrpc": self.rpc_version, "method": method, "id": self.id}
+        if params:
+            payload["params"] = params
+
+        response = requests.get(self.json_rpc_url(payload), headers=self.headers)
+        self.id += 1
+        return json.loads(response.text)
 
     def sendClientCommand(self, command, args=None):
         url = '%s%s' % (self.url(command), utils.joinArgs(args))
@@ -57,6 +71,10 @@ class Client(object):
 
     def url(self, path):
         return 'http://%s:%s/player/%s' % (self.address, self.port, path.lstrip('/'))
+
+    def json_rpc_url(self, payload):
+        url_param = urllib.urlencode({'request': json.dumps(payload)})
+        return 'http://%s:%s/jsonrpc?%s' % (self.address, self.port, url_param)
 
     # Navigation Commands
     def moveUp(self): self.sendCommand('navigation/moveUp')  # noqa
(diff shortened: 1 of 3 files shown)


mjs7231:
From a first glace at this changeset, it looks like you did a lot more than the title states.  Also sending commands to client via RPC, etc. Many of the changes would be welcomed, but I its really hard to merge a branch with multiple unrelated features added.

danielBreitlauch:
It Is only adding functions. It won't break stuff. It contains 3 mayor things. First the rpc call to navigate a client to a certain detail screen. Second, additional objects like the already existing actor e.g. director, genre, rating... And third, functions for all the object to create these objects from a name.

The problem it fixes is to get the plex id of for example an actor just by name. These ids contained in the objects can then be used in search queries.

If you want to see it used have a look at my natural language plex remote project.

mjs7231:
OK, I understand whats going on now and it looks good, thanks. I'll look at merging this one in next after I get the master branch back to a stable state from the audio merge. Things are a bit in flux at the moment form that one.

Did you end up writing a search function on your end that allows you mix and match these available filters?  I'm contemplating changing the search function to be a bit more usable.  This code will give me a huge head start, what would you think about something like the following:

Change the prototype of the Library.search() function:

```python
def search(query, vtype='all', **kwargs):
    # kwargs could be any of the following:
    # 'unwatched': boolean
    # 'duplicates': boolean 
    # 'year': ['1963', ...] 
    # 'decade': ['1960', ...] 
    # 'genre': [list of strings] 
    # 'content_rating': [list of strings]
    # ...you get the idea...
```

The idea here is that search supports everything the current web client supports. Items like `actor` could be passed in as an int or a string and the conversion is done automatically.

danielBreitlauch:
In my use case/library I do the search basically in two steps there. First is the translation of name strings to objects/ids. The second step is your search function that already supports searching/filtering by objects/ids.

The translation works by requesting all items of a kind and comparing the name to the given one, e.g. filtering by genre 'Comedy' would request all genres and then compare their name with 'Comedy'. If it matches the object with its id is returned.
You can insert this step into you search function to make it possible to directly search/filter by strings.

One thing to consider… different use cases may require different compare functions.
For example actors can be compared word wise or by equality. "Bruce Wills" == "Bruce"?.

An additional optimization would be to have the list of all actors, genres etc. cached.

mjs7231:
I really appreciate the contribution here. I'm not going to merge this change directly, but I believe the issues address are largely covered in the master branch.

* Library.Section.Search() now accepts sort and filters in a much more friendly api. A few usage examples are below. One thing I noticed when searching for actors by name is that not all actors are listed in the /library/section/x/actor resultset, it looks like it only shows the top 100 or so. In those cases you'll need to know the id of the actor directly. See the test `test_crazy_search` for a working example.

```
movies = plex.library.section(MOVIE_SECTION)
movies.search(director='Trevorrow')   # search by string.
movies.search(director=['Adam Brooks', 'Adam Green'])   # search by a list of strings (results are OR'd).
movies.search(director=[1234,4567])  # search by ids directly

steven = movies.get('Jurassic Park').directors[0]
movies.search(director=steven)  # search by director object

# Find all movies containing at least one actor from Jurassic Park and directed by steven..
actors = movies.get('Jurassic Park').actors  # note actors is a list
related = movies.search(actors=actors, director=steven)
```

The second part of this merge request is adding RPC command ability to the client.  I was looking at this and had it implemented but decided its just not time to bring that in yet. Currently everything in this API is working off the XML documents, and introducing a new protocol in the mix is a bit confusing. Especially in the client code which I feel still needs some more tlc to get what is currently there working  a bit more solidly.

That said, I did add the new abilities through the XML API. We can now set the audio, language, and navigate to directly to a media object on the client. The only thing I could not figure out is how to call subtitle.next. The way it currently gets set is by getting the subtitle object and calling `client.setSubtitleStream()` with the object directly.

Whew.. that was long. If you have time, play around with the master branch and let me know what else you may need.

danielBreitlauch:
Looks awesome. I will port my lib to use it.
I guess getting around changing subtitles will not be hard. I will try.
The only feature request that now comes to mind… and it is really just a non functional feature, would be caching of library.listChoices() results for speed improvements ;D.

The code is very concise and clean. Good work!

mjs7231:
I agree caching would be nice. I have been thinking about it but neglected to put it in just yet. I was thinking it could be done more globally. If all requests go through server.query() then we can cache everything. The tricky part is knowing how long something should stay in cache and making sure there is a clean easy way to clear and and optionally not use it. We should probably open a discussion in the issues of there isn't one already.

danielBreitlauch:
Concerning the caching… doing it in server.query is probably a good idea. Although keep in mind a lot of the server requests have side effects like playing a movie ;D.

One thing I just saw in your search implementation. The ALLOWED_FILTERS in the shows section is weird. They do not include filters for episode level.
For example:
if you filter shows by year you get all shows that started that year. If you filter episode by year you get the episodes in that year. Other things like 'writer' can only be filtered on episode level, not show level.
And the other way around actors can only be filtered on show level but not episode level.
filterable in episode:  'director', 'writer', 'year', 'decade'
filterable in show: 'actor', 'genre', 'contentRating'

And I see you left filter by actor out completely. Any reason for that?

PR #55, opened 2016-05-04
Adding season and episode number to Season and Episode objects

ttufts:
Added season_number and episode_number by pulling 'index' and 'parentIndex' in as episode_number and season_number respectively. This works for almost all situations, except when all episodes are pulled from a Show object, without going through seasons, the solution here is to check for 'parentIndex' when building the Episode object, if it is missing, the season() function is called and the season_number is pulled in from the Season object. This works in every situation that you can access an Episode or Season object, whether directly from the Show, or from the Season object.

### DIFF
--- a/plexapi/video.py
+++ b/plexapi/video.py
@@ -51,7 +51,7 @@ class Video(PlexPartialObject):
 
     def refresh(self):
         self.server.query('%s/refresh' % self.key, method=self.server.session.put)
-        
+
     def section(self):
         return self.server.library.sectionByID(self.librarySectionID)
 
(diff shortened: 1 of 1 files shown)


mjs7231:
A implemented it slightly differently, but kept the main idea you had here. I realize what originally happened was during the v2 update last month I inadvertently removed the self.index attribute from the Show, Season, and Episode classes which had me confused and broke the old method of getting the season numbers. Anyway, thanks for the bug report and code! Let me know if this change works for you.

commit 36b5d12a192ac0574c33b0849f1087b17f805c9b
Author: Michael Shepanski <mjs7231@gmail.com>
Date:   Thu May 19 23:42:06 2016 -0400
Add back missing index value from Show,Season,Episode; Make season numbers a bit easier to fetch

PR #69, opened 2016-10-19
Adding PlexApiException as a base exception for all other python-plexapi exceptions.

cxhamilton:
I am proposing the addition of a base exception (PlexApiException) that all other python-plexapi exceptions will extend. The purpose of this base exception is for ease of developer use. No need to catch each exception individually, when all you need to know is that at least one exception was raised; and this will provide more meaningful information than catching the built-in Exception.

### DIFF
--- a/plexapi/exceptions.py
+++ b/plexapi/exceptions.py
@@ -4,17 +4,20 @@
 PlexAPI Exceptions
 """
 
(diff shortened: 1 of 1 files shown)


mjs7231:
Good idea. I'll check and merge this after I get back from vacation.

PR #80, opened 2016-12-17
Inital pr for some docs and code 

Hellowlol:
- fixed couble quoted list options
- __setattr__ now uses __dict__ since its twice as fast.
- grab all the data we can, we dont want to reload if we can avoide it.

- autopep

Im not really sure what everything does so i have done some mistakes feel free to correct it.

### DIFF
--- a/plexapi/client.py
+++ b/plexapi/client.py
@@ -4,6 +4,7 @@ PlexAPI Client
 To understand how this works, read this page:
 https://github.com/plexinc/plex-media-player/wiki/Remote-control-API
 """
+
 import requests
 from requests.status_codes import _codes as codes
 from plexapi import BASE_HEADERS, TIMEOUT, log, utils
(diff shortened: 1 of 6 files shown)


Hellowlol:
Oh, i wasnt even close to done :p I was just to get the ball rolling to i could get some feedback :P.

mjs7231:
I looked it over and it seemed fine, mostly doc changes.  Shouldn't affect anything, but I'll run my tests before putting anything new on pip.  Also, I gave you push access.  You've been using this lib for a year and a half now and I haven't really been as active as I'd like myself. :P

mjs7231:
Can I suggest two small things when reformatting code.

1. If the doc string is a one liner, keep the """ triple quotes on one line. """
2. If you are indenting arguments to a function, have the next line only indent 4 spaces and not indent all the way over to the function bracket.

```python
    def this_is_hard_to_follow(baz, bar,
                               rat,
                               goofballs,
                               zippy_ties):

    def this_makes_more_sense(baz, bar,
        rat, goofballs, zippy_ties):
```

Hellowlol:
Thanks for the push access. Yeah, I'll fix that. About the tests any objects if we use pytest? I'll keep the ones we have and make some new.

mjs7231:
No objections at all. It would be very welcome actually.

PR #81, opened 2016-12-20
Fix stuff

Hellowlol:
- Fix plex.library.all()
- Add transcode url. Dunno if we should handle the http too.
- Try to fix NA for thumbs.


Hellowlol:
https://github.com/mjs7231/python-plexapi/issues/75

PR #86, opened 2016-12-30
Better find/filter

Hellowlol:
This is a just a suggestion to allow lowerlevel access to findItem. And a general helper to filter on things.

Possible use could be directly on a Element
```
def bf():
    # get season 1 ep 1
    d = {'index': '1', 'parentIndex': '1'}

    rating = '111028'

    path = '/library/metadata/%s/allLeaves' % rating

    x = betterFind(plex, path, typ='all', **d)
    print x
```

Or to filter on the attrs of media objects

```
list(filterAll(section.all(), typ='all', **{'rating<': 7, 'genres': 'comedy'}
```

### DIFF
--- a/plexapi/utils.py
+++ b/plexapi/utils.py
@@ -7,8 +7,11 @@ Attributes:
     NA (TYPE): Description
     SEARCHTYPES (TYPE): Description
 """
+import inspect
 import re
 from datetime import datetime
+from xml.etree import ElementTree as ET
+
 from plexapi.compat import quote, urlencode
 from plexapi.exceptions import NotFound, UnknownType, Unsupported
 from threading import Thread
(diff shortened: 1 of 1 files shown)


Hellowlol:
@mjs7231 What your take on this?

mjs7231:
Just catching up now, I don't quite understand, but it's late and I probably just need sleep.  I'll take a closer look tomorrow. :D

mjs7231:
Had a chance to look at this one.  I think we need to think about it a bit more.  The idea behind it seems solid, but there are a few things I'm unsure of.

Red flags are always raised for me when I see inspect imported in production code. To me that generally means there is a better approach, or design flaw that just needs more thought.

Second, I'm not so keen on implementing a new unique syntax for searching. The rating< rating> thing is very weird to me. I feel like we could use a lot more examples of possible use cases before this should be considered. When looking at the movie filter and sorting options currently available in the PlexWeb client, it seems like this would be useful for numeric or date options such as `year`, `rating`, `duration`, `date added`, `date released`, `date watched`.

Because of the weird API and complexity it introduces, im still wondering if this is something in the scope of this API project , or better left as something the user should implement for their specific case outside the API.

Hellowlol:
The main user case is speed. The only issue I have with plexapi that just insanely slow compare to working json directly as I'm used to. My goal with plexapi is to get it fast enough to I can replace my json in htpc manager.

This only a issue when iterating over large data sets (like we typically do to find shit).
https://github.com/mjs7231/python-plexapi/pull/90/ include some changes to cast since I noticed we where using 5,8 sec just casing int/bool on stuff. 

Plex handles most of the weird < on numerical values already but not all of them. (Not supported in plexapi ATM)

I'll post some use cases and code examples later.
Find any element with rating over 8 and where one of the actors where Wesley Snipes and was directed by Spielberg 

On my phone now so I'll clean up this post later..

mjs7231:
> The main user case is speed. This only a issue when iterating over large data sets (like we typically do to find shit). #90/ include some changes to cast since I noticed we where using 5,8 sec just casing int/bool on stuff.

I apologize. I totally missed the fact that you are filtering _before_ building out the Python object. That would definetly make things much faster.

One place I have seen filters used in kwargs is Django, and I quite like the approach they used, maybe we can steal it? Rather than using the special character <>!, you can append __lt, __gt, __ne to the of the filter argument. It allows us to not force kwargs into a dictionary which looks cleaner. So your example would look more like this:

```python
list(filter_all(section.all(), type='all', rating__lt=7, genres=comedy))
```

Hellowlol:
I used the > inside the dict because that is what plex uses. The idea is to use plex for the things for pop kwargs and use the filter to handle the things plex don't support with the same interface (same reason it supports both elements and attrs)

One drawback is that all elements attrs are strings :(.

So it still needs some work.

mjs7231:
> I used the > inside the dict because that is what plex uses.

You're diving int an area of Plex I clearly don't know much about. If that's the case, then its a strong argument to keep it the way you have it. I guess its a debate between requiring you push your kwargs into a dict and use **{} which looks messy, or try to keep things closer to Plex syntax.

Where does Plex allow filtering less then or greater than?

Hellowlol:
> Where does Plex allow filtering less then or greater than?

http://ip:port/library/all?rating>=8&genre=235&X-Plex-Token=TOKEN

mjs7231:
That's cool, didn't know that! -- Thinking about the API, I'm still leaning toward the syntax of __lt, __lte notation.  Something about it just seems much cleaner and Pythonic to me from a user of the API's perspective.

I do have another question. -- Why remove the > bit from the kwarg and filter the list in Python land if it can be passed directly to to Plex's API as part of the URL and presumably be filtered by the server?

This is a welcome change and am excited to have it merged once we flesh it out! :D

Hellowlol:
> I do have another question. -- Why remove the > bit from the kwarg and filter the list in Python land if it can be passed directly to to Plex's API as part of the URL and presumably be filtered by the server?

Because we are using the real_k to figure out how to and when to check <=s but the real attributes in the class or element dont have them.

If you wanna keep __lt__ etc you have to do that part yourself, i have no idea how to handle this :D and we lose the compability with how plex handles this.

Hellowlol:
> One drawback is that all elements attrs are strings :(.

>So it still needs some work.

On second though i think thats just fine. FilterAll is support to be a helper for lowlevel access not something the user would be exposed to.

mjs7231:
This functionality is covered on the 3.0.0 branch.

PR #87, opened 2016-12-31
cast viewCount to int

WyseNynja:
I ran into this:

```
  File "/pyenv/lib/python3.5/site-packages/plexapi/video.py", line 158, in unwatched
    return self.episodes(watched=False)
  File "/pyenv/lib/python3.5/site-packages/plexapi/video.py", line 148, in episodes
    return utils.listItems(self.server, leavesKey, watched=watched)
  File "/pyenv/lib/python3.5/site-packages/plexapi/utils.py", line 235, in listItems
    if watched is False and elem.attrib.get('viewCount', 0) >= 1: continue
TypeError: unorderable types: str() >= int()
```

Simply casting to an int should fix.

### DIFF
--- a/plexapi/utils.py
+++ b/plexapi/utils.py
@@ -460,9 +460,9 @@ def listItems(server, path, libtype=None, watched=None, bytag=False):
     for elem in server.query(path):
         if libtype and elem.attrib.get('type') != libtype:
             continue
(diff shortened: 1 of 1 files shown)


mjs7231:
This is probably fine in this case, but if we do hit an issue we may need to use the utils/cast() function as it won't hard fail if the value for viewCount is blank for whatever reason.

PR #88, opened 2017-01-02
Fix all, fix thumb.

Hellowlol:


plexapi/video.py
```diff
@@ -169,11 +170,11 @@ def _loadData(self, data):
         self.viewedLeafCount = utils.cast(
             int, data.attrib.get('viewedLeafCount', NA))
         self.year = utils.cast(int, data.attrib.get('year', NA))
-        if self.isFullObject():
-            self.genres = [media.Genre(self.server, e)
-                           for e in data if e.tag == media.Genre.TYPE]
-            self.roles = [media.Role(self.server, e)
-                          for e in data if e.tag == media.Role.TYPE]
+        #if self.isFullObject(): # will be fixed with docs.
```
mjs7231: I think we should discuss this one a bit more.  If I remember correctly, the genres and roles were only listing a subset of the list on the search list which is why I was checking for a full object before. As there is currently no way to ask for a full update (or update) the partial list once it's been set, this may have broke something.

Hellowlol: Thanks! I havnt thought about that. I really should start on some tests. Im not sure the best way to to patch the http responses from pms.

PR #90, opened 2017-01-03
Inital hub.

Hellowlol:
plex.search now uses hubs behind the curtains.


mjs7231:
I'm very excited for this one! -- I tried to implement it myself a few months back, but ended up ditching my approach at the time. I think we should definetly include some tests for this one before we merge to make sure it's solid now as well as keep up on possible degradations in the future.

Hellowlol:
It should be ok, i tested manually but I would feel a lot better if I could run some automated tests. I was thinking I could recorded some http responses so we can get the ball rolling.

PR #91, opened 2017-01-04
Suggestion for episode

Hellowlol:
Possible uses
````
show = show.search('The Blacklist')[0]

show.episode(season=1, episode=1)
<Episode:116263:The.Freelancer>

show.season(1)
<Season:116260/children:Season.1>
````
Im not sure about the api.

### DIFF
--- a/plexapi/video.py
+++ b/plexapi/video.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 
 from plexapi import media, utils
+from plexapi.exceptions import NotFound
 from plexapi.utils import Playable, PlexPartialObject
 
 NA = utils.NA
@@ -195,9 +196,8 @@ class Show(Video):
         Args:
             title (str, int): fx Season 1
         """
-        title = str(title)
-        if title.isdigit():
-            title = 'season %s' % title
+        if isinstance(title, int):
+            title = 'Season %s' % title
 
         path = '/library/metadata/%s/children' % self.ratingKey
         return utils.findItem(self.server, path, title)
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: plexapi/video.py [hunk 1]
```diff
@@ -189,12 +189,16 @@ def seasons(self):
...
 
-    def season(self, title):
+    def season(self, title=None):
         """Returns a Season
 
         Args:
-            title (str): fx Season1
+            title (str, int): fx Season 1
         """
+        title = str(title)
```
mjs7231: I think using isinstance(title, int) might be a better choice here than converting it to a string then checking isdigit()

plexapi/video.py
```diff
@@ -207,9 +211,18 @@ def episodes(self, watched=None):
         leavesKey = '/library/metadata/%s/allLeaves' % self.ratingKey
         return utils.listItems(self.server, leavesKey, watched=watched)
 
-    def episode(self, title):
-        path = '/library/metadata/%s/allLeaves' % self.ratingKey
-        return utils.findItem(self.server, path, title)
+    def episode(self, title=None, season=None, episode=None):
```
mjs7231: I like the idea of title or season & episode. We should raise an exception at the end of this function if we fall through the if statements explaining that they need to specify both season & episode for the second case to work.

plexapi/video.py
```diff
@@ -268,25 +281,38 @@ def episodes(self, watched=None):
...
 
-        
+
         """
         childrenKey = '/library/metadata/%s/children' % self.ratingKey
         return utils.listItems(self.server, childrenKey, watched=watched)
 
-    def episode(self, title):
-        """Find a episode with a matching title.
+    def episode(self, title=None, season=None, episode=None):
```
mjs7231: Keeping season as an argument here seems incorrect, probably should be removed.

plexapi/video.py
Hellowlol: Agree. Should we raise not found or just return None if we don't find any episode too?

plexapi/video.py
Hellowlol: i wasn't sure if I should add season, I only did so episode in show and season accept the same kwargs. I'll remove it

plexapi/video.py
Hellowlol: Yep

PR #93, opened 2017-01-12
i think i got a list with [None]

WyseNynja:
While trying to connect to a friend's server that is currently offline, I got this exception

```

...

  File "/code/playlists/models.py", line 106, in import_servers
    conn = plex_server_resource.connect()
  File "/pyenv/src/plexapi/plexapi/myplex.py", line 335, in connect
    results[0].baseurl, results[0].token)
AttributeError: 'NoneType' object has no attribute 'baseurl'
```

I believe this happened because `r[2]` somehow ended up as None and so the filtering wasn't enough


Hellowlol:
Ah, my bad. if should be ```results = [r[2] for r in results if r and r[2] is not None]```

PR #115, opened 2017-02-12
Kill transcode

Hellowlol:
And other fixes. This will require a review.

### DIFF
--- a/plexapi/client.py
+++ b/plexapi/client.py
@@ -47,8 +47,12 @@ class PlexClient(PlexObject):
     key = '/resources'
 
     def __init__(self, baseurl, token=None, session=None, server=None, data=None):
+
         self._baseurl = (baseurl or CONFIG.get('authentication.client_baseurl')).strip('/')
         self._token = token or CONFIG.get('authentication.client_token')
+        if server and server._token and not self._token:
+            self._token = server._token
+
         if self._token:
             logfilter.add_secret(self._token)
         self._server = server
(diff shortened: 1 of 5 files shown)


Hellowlol:
We need a landscape file we so it stops complaining about private access.

mjs7231:
I think we overlapped a bit of our efforts with the PlexClient issue. What are the things we should be keeping here? The kill-transcode should def by copied in.

It also looks like you found a few missing attributes in PlexClient. I think we should probably not be overriding the names of attributes to new names as in this example.  I used to do that, and things get pretty confusing pretty quickly. (`self.protocolCapabilities = data.attrib.get('protocolCapabilities', data.attrib.get('provides', '')).split(','))`)

The playQueue changes I don't fully understand, but I'm also tired (its 2am here). What happens if we call `playMedia()` with both media and playQueue specified? If that works right, what happens if media is not in the playQueue items? -- I was thinking we could have a `playlist.toPlayQueue()`, then a new function next to playMedia called something like startPlayQueue() (that playMedia would call as well).

How is `_find_media_type()` used?

Hellowlol:
> How is _find_media_type() used?
Atm its not, its just a nice helper to use if your lazy and want to figure out what mtype to pass automatically. But its rather slow, so i was thinking we should just change mtype to mtype='video' since that is what pms is generally used for and require the user only to set the correct mtype if the player isnt playing video.

> The playQueue changes I don't fully understand, but I'm also tired (its 2am here). What happens if we call playMedia() with both media and playQueue specified? 
If playQueue is passed, that gets presedence and it used if not we autocreate it for the user has we have done in the past.

Hellowlol:
Im gonna close and split this PR. I never intended for it to do so much. 
I'll start with making clients better first.

mjs7231:
for playMedia(), I think you can skip the argument all together, and just pass in a `playMedia(media=<playQueue>` and do the right thing.  Pretty much the same thing you already did, but drops an argument?

Hellowlol:
That would be the best, I believe it's possible to do this will some small changes to the createplayque.

Hellowlol:
I know this API is intended to mimic the xml attrs but sometimes this feels really awkward. Say I want the name of a server. That's friendlyname. A client is name, and there are a couple of other cases. Would it be ok if we added some aliases so it's a bit  more intuitive to use?

mjs7231:
It might be fine. I just found that when I renamed Plex attrs in the API before it was generally because I was doing something awkward or not really intended. With clients, I question the usefulness of listing sessions then trying to connect to the client from the session list. That would only work if they owned the client that was already playing the media. And if they own the client then connecting from the devices list seems like the better path anyway.

Anyway, if you guys really think it's needed I don't want to stand in the way. Just seems like an odd path to me.

PR #141, opened 2017-04-25
Fixed regression involving invalid token requirement

centicosm:
I found what appear to be two small regressions in client.py and server.py when compared to the 2.0 branch.  The url function in client.py requires the optional token preventing tokenless operation.  This pull request reverts the behavior to allow the token to be optional once more.  In addition, the client function in server.py was updated to explicitly name the baseurl parameter as it is no longer the first argument of the constructor.

### DIFF
--- a/plexapi/client.py
+++ b/plexapi/client.py
@@ -171,8 +171,8 @@ class PlexClient(PlexObject):
 
     def url(self, key):
         """ Build a URL string with proper token argument. """
-        if not self._baseurl or not self._token:
-            raise BadRequest('PlexClient object missing baseurl or token.')
+        if not self._baseurl:
+            raise BadRequest('PlexClient object missing baseurl.')
         if self._token:
             delim = '&' if '?' in key else '?'
             return '%s%s%sX-Plex-Token=%s' % (self._baseurl, key, delim, self._token)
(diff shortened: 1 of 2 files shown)


pkkid:
Thanks, good finds.

Can I ask what the use case is for calling `url()` without a token specified? I fear it may cause confusion if they try to use other functionality that will result in permission denied if a token is required and we don't call it out explicitly. Maybe there is a way to support both use cases?

Hellowlol:
If he has disabled auth?

centicosm:
My Plex setup is only handling devices on my local network so auth is disabled to remove the dependency on having/using a Plex account (which, as far as I can tell, is required to get the token).  In the current version, a client can't playMedia without the token.  I noticed the pypi documentation and current documentation differ in that the example of getting a plex server instance without a token is no longer listed.  Is the library moving away from such a use case?

pkkid:
We're not moving away from anything. Our goal is to support everything the PlexWeb client can do. I'll have to try connecting with disabled auth again, I was having no luck getting it working after an update around Jan.

Thanks for the fixes, merging them now.

pkkid:
Also note: The pypi docs are different. We changed the API slightly to allow using a base class for devices and server to help keep consistency across the library. There will be much more documentation moving forward, and we have a good start on a test suite to prevent regressions.

PR #151, opened 2017-07-16
Init edit.

Hellowlol:


Hellowlol:
Still need to add tests etc. I just wanted some feedback.

JonnyWong16:
An issue I see is that if you only want to add a single collection/label/genre, it will overwrite the previous collection/label/genre. Also, it will remove all collections/labels/genres.

Hellowlol:
@JonnyWong16  is the - to remove for labels/genres etc all of them or just collections?

JonnyWong16:
Looks like it's the same for labels and genres
```
genre[].tag.tag-
label[].tag.tag-
collection[].tag.tag-
```

Hellowlol:
Awesome, Thanks.

plexapi/base.py
```diff
@@ -314,6 +316,75 @@ def isPartialObject(self):
...
+                locked (bool): lock this field.
+                remove (bool): If this is active remove the tags in items.
+            Returns:
+                None
+
+        """
+        if not isinstance(items, list):
+            items = [items]
+
+        existing_cols = [t.tag for t in self.collections if t and remove is False]
```
JonnyWong16: Always `self.collections` :stuck_out_tongue:

plexapi/base.py
```diff
@@ -314,6 +316,75 @@ def isPartialObject(self):
...
+           Returns:
+                None
+        """
+        self._edit_tags('collection', collections)
+
+    def removeCollection(self, collections):
+        self._edit_tags('collection', collections, remove=True)
+
+    def addLabel(self, labels):
+        self._edit_tags('collection', labels)
```
JonnyWong16: `'label'`

JonnyWong16:
Needs to reassign the existing collections when removing.

```
# Removing some collection
"collection[0].tag.tag": "Collection 2"  <-- "Collection 2" is now at index 0
"collection[].tag.tag-": "Collection 1,Collection 3"
```

Hellowlol:
> Needs to reassign the existing collections when removing.

# Removing some collection
"collection[0].tag.tag": "Collection 2"  <-- "Collection 2" is now at index 0
"collection[].tag.tag-": "Collection 1,Collection 3"

I didnt understand what you mean. if remove support remove by value why should be care about the index, plex should handle this. Thanks for the feedback btw.

JonnyWong16:
I'm just following what inspecting the API call in Plex Web gives. It shuffles all the remaining collections down to index [0].

plexapi/utils.py
```diff
@@ -277,3 +277,23 @@ def download(url, filename=None, savepath=None, session=None, chunksize=4024, un
...
     # finished; return fillpath
     return fullpath
+
+
+def tag_helper(tag, items, locked=True, remove=False):
+    """Simple tag helper for editing a object."""
+    if not isinstance(items, list):
+        items = [items]
+
+    d = {}
```
JonnyWong16: Make `remove` a `list` instead of a `bool`?

```
d = {}

if remove:
    for item in remove:
        try:
            items.remove(item)
        except ValueError:
            pass

    tag_name = '%s[].tag.tag-' % tag
    d[tag_name] = ','.join(remove)

for i, item in enumerate(items):
    tag_name = '%s[%s].tag.tag' % (tag, i)
    d[tag_name] = item

d['%s.locked' % tag] = 1 if locked else 0

return d
````

Hellowlol:
I think we already supports this, however i don't think i'm gonna bother adding a method for every change. 

````
{'titleSort.value': 'oh please',
 'titleSort.locked: '1'}
````

plexapi/utils.py
Hellowlol: cba, it works with what we have today.

JonnyWong16: I'm not sure if there will be any negative effects for not setting the indices back to 0. What happens after you edit the tags with the Python API, then go an edit the tags in Plex Web?

Hellowlol: Nothing, it just works.

PR #152, opened 2017-07-17
Shared server

Hellowlol:
Add a simple way to login to your own server as another users. This is used to create playlists that can be shared.

### DIFF
--- a/plexapi/myplex.py
+++ b/plexapi/myplex.py
@@ -280,7 +280,6 @@ class MyPlexAccount(PlexObject):
         requested = [MyPlexUser(self, elem, self.REQUESTED) for elem in self.query(self.REQUESTED)]
         return friends + requested
 
-
     def _getSectionIds(self, server, sections):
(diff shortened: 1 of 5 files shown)


Hellowlol:
This works. I just had to find the test auth tolken and url.

Hellowlol:
Closing this, i want to see if cant use somthing that already exists. Some fixes here is needed for other stuff. So dont delete it.

plexapi/myplex.py
```diff
@@ -380,6 +386,15 @@ def _loadData(self, data):
         self.username = data.attrib.get('username')
         self.servers = self.findItems(data, MyPlexServerShare)
 
+    def get_token(self, machineIdentifier):
+        try:
+            for item in self._server.query(self._server.FRIENDINVITE.format(machineId=machineIdentifier)):
+                print(item.attrib.get('userID'), self.id)
```
Hellowlol: It should be removed.

Issue #158, opened 2017-07-25
plexapi.exceptions.BadRequest: (406) not_acceptable


UnR34L:
#Plex Media Server Version : 1.7.6.4058
OS: FreeBSD 11.0-STABLE
Python: 2.7.13
pip 9.0.1

websocket-client and requests are installed.

```
#!/usr/local/bin/python2.7

from plexapi.server import PlexServer
baseurl = 'http://192.168.0.54:32400'
token = 'TOKEN'
plex = PlexServer(baseurl, token)

ozzy = plex.library.section('Trailers').get('Ozzy')
client = plex.client("DESKTOP-B88TLB1")
client.playMedia(ozzy)
```

**Traceback**

`Traceback (most recent call last):
  File "./test.py", line 10, in <module>
    client.playMedia(ozzy)
  File "/usr/local/lib/python2.7/site-packages/plexapi/client.py", line 441, in playMedia
    }, **params))
  File "/usr/local/lib/python2.7/site-packages/plexapi/client.py", line 181, in sendCommand
    return self.query(key, headers=headers)
  File "/usr/local/lib/python2.7/site-packages/plexapi/client.py", line 153, in query
    raise BadRequest('(%s) %s; %s' % (response.status_code, codename, errtext))
plexapi.exceptions.BadRequest: (406) not_acceptable;`

Hellowlol:
What is the client your trying to play on?

UnR34L:
With the list of clients I get from 

for client in plex.clients():
print(client.title)


I tried 'DESKTOP-B88TLB1' which should be Plex Media Player then I tried 'Plex Web (Chrome)' and a Samsung Smart TV but that gives the error 

`plexapi.exceptions.BadRequest: (401) unauthorized; <html><head><title>Unauthorized</title></head><body><h1>401 Unauthorized</h1></body></html>`

It says unauthorized so I tried the logging into MyPlex method, which gave the same error, and I know the username and password are correct.

UnR34L:
Tried a Android client, it plays but gives the error

`Traceback (most recent call last):
  File "./test.py", line 9, in <module>
    client.playMedia(alien)
  File "/usr/local/lib/python2.7/site-packages/plexapi/client.py", line 441, in playMedia
    }, **params))
  File "/usr/local/lib/python2.7/site-packages/plexapi/client.py", line 181, in sendCommand
    return self.query(key, headers=headers)
[...]

PR #172, opened 2017-08-15
Fixed two bugs

itssimon:
Made MyPlexDevice.connect() return a Client object (return statement was missing) and removed timeline subscription when playing media using PlexClient.playMedia() because this makes RasPlex (OpenPHT 1.8.0) crash.

### DIFF
--- a/plexapi/client.py
+++ b/plexapi/client.py
@@ -429,13 +429,14 @@ class PlexClient(PlexObject):
         if not self._server:
             raise Unsupported('A server must be specified before using this command.')
         server_url = media._server._baseurl.split(':')
-        try:
-            self.sendCommand('timeline/subscribe', port=server_url[1].strip('/'), protocol='http')
-        except:
+        # Don't subscribe to timeline because this makes RasPlex crash
+		#try:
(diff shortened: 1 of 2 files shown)


pkkid:
> Removed timeline subscription when playing media using PlexClient.playMedia() because this makes RasPlex (OpenPHT 1.8.0) crash.

This is interesting. We added that because playMedia() was not working on some clients without it. We might have to find a better way to know when to use that and when to not or at the very least include an optional argument for it. Issue #158 from 3 weeks ago is when we added it.

Let me look into this a bit more before we merge this one. -- I'll  get the return statement added sooner.

Hellowlol:
This isnt a proper fix we cant just disable timeline call as it require by other clients. We need to disable this just for openpht. I do find it strange that openpht don't handle this better.

itssimon:
Agree @Hellowlol , that's probably the way to go. I mistakenly called this a fix because it fixed it for my setup.

pkkid:
Not going to merge this one as we are fixing OpenPHT in another way, however I still need to cherry pick the other fix you had in here.

itssimon:
No worries.

Hellowlol:
Open pht should now btw.

PR #176, opened 2017-08-18
Fix for openpht and fix failing tests

Hellowlol:


pkkid:
Im going to change this up slightly and start a dict of ClientName => Metadata about client to make things work. Hard to explain, but I think its a cleaner approach than having if statements calling out specific clients in the code.

Hellowlol:
Alright.

PR #184, opened 2017-09-08
Home user and remove sections

blacktwin:
def updateFriend() update to allow for removing sections if `remove_sections=True`. 

def user() update to find Home user names.

### DIFF
--- a/plexapi/myplex.py
+++ b/plexapi/myplex.py
@@ -216,7 +216,7 @@ class MyPlexAccount(PlexObject):
         url = url.format(userId=user.id)
         return self.query(url, self._session.delete)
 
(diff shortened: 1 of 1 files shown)


blacktwin:
Didn't test enough. Unable to add shares after remove_sections=True.

Hellowlol:
I'd prefere if remove section is true. Since a lot of things are thruthy. Can you update the doc string?

pkkid:
Also, to stick with our camelCase formating it should be removeSections instead of remove_sections.

pkkid:
@Hellowlol Are you good with this one? I don't quite understand this area enough.

Hellowlol:
Honestly i cant even remember there being a update friend method. It looks like my code but i cant remember the name. As long as it works ill say merge it and ill go back and add a test for this later.

PR #187, opened 2017-09-16
Update video.py

nitemare:
Adding title and parentTitle to Episode variables

### DIFF
--- a/plexapi/video.py
+++ b/plexapi/video.py
@@ -507,6 +507,10 @@ class Episode(Video, Playable):
         if self._seasonNumber is None:
             self._seasonNumber = self.parentIndex if self.parentIndex else self.season().seasonNumber
         return utils.cast(int, self._seasonNumber)
+		
+    def seasonEpisode(self):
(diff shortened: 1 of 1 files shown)


plexapi/video.py
```diff
@@ -503,6 +507,10 @@ def seasonNumber(self):
         if self._seasonNumber is None:
             self._seasonNumber = self.parentIndex if self.parentIndex else self.season().seasonNumber
         return utils.cast(int, self._seasonNumber)
+		
+    def seasonEpisode(self):
```
pkkid: Is there a difference between this function and the property right below it?

nitemare: oh, no, that would be my mistake adding it here, i just installed and started using the Plexapi, i installed it with the 'pip install plexapi' command, but this function was not included in in the video.py file. I guess i assumed it would have installed the latest version, and the first change i proposed wasn't there so overlooked the function was already in the github file here. i have no explanation.

pkkid:
No problem. I figure that's that happened.

PR #189, opened 2017-09-26
Better cli

Hellowlol:


Hellowlol:
@pkkid What's your take one this one. 
I wanted to add the ability to select several eps at the same time. 
and a console entry point.

pkkid:
It doesn't seem to be working for me. I tried to download a file, and it just exited without downloading after choosing a server and movie. Second attempt, it exited immediately after asking what I'm looking for, I assume it found no results even tho the old version did.

A have a few comments anyway:

- I consider the tools directory as a playground. It's a good place for us to experiment as well as provide anything useful we create with others. With that in mind, I think we can be much less strict with what we allow in here (working or in-progress).
- I like that you took `showstatus` out of the official download function. It felt really dirty that I put it there.
- I would generally avoid using one character variable names unless its a one-liner list comprehension, etc.


Run 1:
```bash
>> ./plex-download.py 
Authenticating with Plex.tv as pkkid..

  0: pkkid-home
  1: homeserv
  2: pkkid-plexapi

Choose a Server: 2
What are you looking for?: foo

  0: (Show) The 100
  1: (Movie) Big Buck Bunny

Choose result: 1
```

Run 2:
```bash
>> ./plex-download.py 
Authenticating with Plex.tv as pkkid..

  0: pkkid-home
  1: homeserv
  2: pkkid-plexapi

Choose a Server: 2
What are you looking for?: Big
```

Hellowlol:
It worked fine here. I'll fix the bugs drop the part if a cli. I'll make a seperate package for that and release it on pypi. My mother in law sometimes uses my plex credentials to login to plex to download stuff. It would be nice to have a better tool for that

Hellowlol:
btw, regarding plex cli. I have created a small cli tool it pre pre alpha but it works better then this pr i think. You can find it here https://github.com/Hellowlol/plexcli

pkkid:
This is working great for me now. :+1:

Issue #191, opened 2017-09-28
Bad type conversion if movie is unwatched and hasattr called


tdorsey:
Refs #190 
```python
  File "/home/tommy/plexsync/base.py", line 16, in dump
    if hasattr(obj, attr):
  File "/usr/lib/python3.6/site-packages/plexapi/base.py", line 274, in __getattribute__
    value = utils.getattributeOrNone(PlexPartialObject, self, attr)
  File "/usr/lib/python3.6/site-packages/plexapi/utils.py", line 81, in getattributeOrNone
    return super(obj, self).__getattribute__(attr)
  File "/usr/lib/python3.6/site-packages/plexapi/video.py", line 49, in isWatched
    return bool(self.viewCount > 0)
*TypeError: '>' not supported between instances of 'NoneType' and 'int'*
```

On an unwatched movie, viewCount is None and the conversion fails.

Hellowlol:
It should be None unless the object isnt loaded.

See https://github.com/pkkid/python-plexapi/blob/master/plexapi/video.py#L42

pkkid:
You you break the code snippet to it's simplest form and post it here so we can understand what you were doing?

If you're still looking at converting these to json, you may want to take a look and understand [plex-listattrs.py](https://github.com/pkkid/python-plexapi/blob/master/tools/plex-listattrs.py). I essentially did a recursive walk of the objects in there. It gets tricky when trying to deal with partial objects returned in search listing and such.

tdorsey:
Yeah, I’ll get a sample together. Essentially I’ve overridden the default json encoder for anything inheriting from PlexObject and am naively json dumping its __dict__, but that’s causing issues with recursing into things like the session state of a PlexServer object.

Hellowlol:
I took a look at your project and i see some things that we can do to speed up your querys.
ATM you do
````
def getMedia(server, section):
    results = server.library.section(section).search()
    api_objects = []
    if section == "Movies":
        for r in results:
            a = APIObject(r)
            api_objects.append(a)
        return set(api_objects)
    elif section == "TV Shows":
        # Show objects don't have a GUID, so grab the first episode and read it from there
        # This is really inefficient, since to query a single episode, the python-plexapi creates a
        # query for all episodes, and then filters down to one
        for r in results:
            episode = next(iter(r.episodes() or []), None)
            a = APIObject(episode)
            api_objects.append(a)
        return set(api_objects)
````

Thats really slow.

````
[...]

PR #192, opened 2017-09-28
Fix TypeError when viewCount is not set

### LINKED ISSUE #190: Objects are not JSON serializable
server = PlexServer()

json.dumps(server) fails because Servers contain Libraries, which contain Sections, etc. 

The easiest thing to do may be to implement custom json encoder/decoders for PlexObject

tdorsey:
Fixes #190

### DIFF
--- a/plexapi/video.py
+++ b/plexapi/video.py
@@ -44,8 +44,11 @@ class Video(PlexPartialObject):
     @property
     def isWatched(self):
         """ Returns True if this video is watched. """
-        return bool(self.viewCount > 0)
-
+        if not self.viewCount:
+            return False
+        else:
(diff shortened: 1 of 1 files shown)


tdorsey:
This may actually be a symptom of a bigger issue when retrieving element data. _loadData is doing the following: 

''

  self.viewCount = utils.cast(int,data.attrib.get('viewCount', 0))

''

I assume it’s meant to already be returning 0 if it’s not set, but that call is returning None and so we see the exception in isWatched. I assume similar errors might crop up for any ints plex doesn’t return

tdorsey:
This actually fixes #191, and refs #190.

PR #194, opened 2017-09-29
download improvements

Hellowlol:
- Add missing showstatus doc
- add name and total to tqdm so the bar is displayed properly


pkkid:
looks good to me.

PR #195, opened 2017-09-30
add broken delete to parts.

Hellowlol:
So this should work but it dont. Can you spot why? the only difference i can notice is that plex web uses the "local external ip"


Hellowlol:
got it, its the media id not the partid.

PR #203, opened 2017-10-05
Dunders

Hellowlol:


pkkid:
Everything in Plex that uses the `/library/metadata/<key>` scheme, has a unique identifier already.  Would it be worth using that key as the hash id? It seems simpler and you could but the function in a higher level base class to get the method on everything across the board.

Hellowlol:
The key is only unique for that server. So we hash against repr so we can use it against other library's from other servers

pkkid:
Cool idea. If that’s a goal, we shouldn’t be using repr as it contains the key. Maybe some concatenation of type, title, and year.

Hellowlol:
Yes but it also contains the name it unlikely that both the key and the name is the same in both objects. Year would be nice.

Issue #206, opened 2017-10-07
Missing attributes (that dont exist at all) should raise AttributeError


Hellowlol:
I was puzzled when i found out that plexapi returns None for all missing attributes. 

What i tested this with.
````python
movie = pms.search('16 blocks')[0]
# ask for a guid so it reloads automatically
print(movie.guid)
# This should raise a AttributeError
print(movie.asshat)
````

My naive fix was
````python
    def __getattribute__(self, attr):
        value = super(PlexPartialObject, self).__getattribute__(attr)
        # Check a few cases where we dont want to reload
        if attr == 'key' or attr.startswith('_'): return value
        if value not in (None, []): return value
        if self.isFullObject(): return value
        # Log warning that were reloading the object
        clsname = self.__class__.__name__
        title = self.__dict__.get('title', self.__dict__.get('name'))
        objname = "%s '%s'" % (clsname, title) if title else clsname
        log.warning("Reloading %s for attr '%s'" % (objname, attr))
        # Reload and return the value
        self.reload()
        return super(PlexPartialObject, self).__getattribute__(attr)
````

I havnt tested anything else.

pkkid:
This should be closed, correct?

Hellowlol:
Yes. I just forgot

PR #207, opened 2017-10-07
missing attrs should raise attribute errors.

Hellowlol:
This just so travis test the pr.

### DIFF
--- a/plexapi/audio.py
+++ b/plexapi/audio.py
@@ -78,6 +78,10 @@ class Artist(Audio):
         self.similar = self.findItems(data, media.Similar)
         self.collections = self.findItems(data, media.Collection)
 
+    def __iter__(self):
+        for album in self.albums():
+            yield album
+
     def album(self, title):
         """ Returns the :class:`~plexapi.audio.Album` that matches the specified title.
 
@@ -150,6 +154,10 @@ class Album(Audio):
     """
     TAG = 'Directory'
     TYPE = 'album'
+
+    def __iter__(self):
+        for track in self.tracks:
+            yield track
 
     def _loadData(self, data):
         """ Load attribute values from Plex XML response. """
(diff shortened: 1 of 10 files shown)


Hellowlol:
https://github.com/pkkid/python-plexapi/issues/206

Hellowlol:
hmm, seems like we should a bug in the tests. As far as i can see there is no .player but since every missing attributes has returned None this has passed.

Hellowlol:
I'll fix the failing tests and merge this 👍

pkkid:
Don't merge this yet. It's a very touchy part of the code and I want to take a look at it. For example: I think player is only an attribute in some cases (ie: actively playing sessions).

Hellowlol:
Alright, I'll wait for your approval. I think your wrong. I can only find players. I think player is a typo that has been copy pasted around in the tests

pkkid:
You're probably right then. I'm going off memory. This is a great improvement, I'm just nervous because as you mentioned before.. dragons.

Hellowlol:
Have you had time to test this?

pkkid:
I haven't. It's going to be a bit as things are pretty busy at work these weeks. -- How did you handle the case when viewCount is zero it's not actually an attribute on the item?

Hellowlol:
I havnet really thought about about viewCount but since we hate using .get"viewCount", 0) it should always have a value. I think one of ous test would have failed nevertheless I'm gonna make a test just for that and fix the .player

pkkid:
If we have to write special code to make viewCount work, it makes me wonder if there are other attributes similar to it that hide when equal to 0, false, inactive, etc. -- If it comes to it, we may be can account for missing attrs in the utils.cast function; its a good place where we know we want to populate an attribute, weather we see it in the data or not.

Hellowlol:
Seems to be working fine now 👍

pkkid:
:+1:

PR #219, opened 2017-10-22
Lets see if this is enough.

Hellowlol:


pkkid:
It should be enough until the next app comes along saying there is a security risk because they log the HTTP headers. :P

Hellowlol:
Restarted the build, got green after plex fixed there issues.

PR #220, opened 2017-10-24
add MyPlexUserShares class

blacktwin:
I wanted to be able to see user shares from my server. 

`plex.myPlexAccount().users_shares(plex)[0].__dict__` will show data from https://plex.tv/api/servers/{machineId}/shared_servers.
seems redundant to use server object twice. Not sure how to work around that.

`plex.myPlexAccount().users_shares(plex)[0].shared_sections` will display shared sections in listed dict with section id, key, and name

```python 
print(plex.myPlexAccount().users_shares(plex)[0].shared_sections)

[   {   'id': 'xxxx', 'key': '16', 'name': 'Audio Books'},
    {   'id': 'xxxx', 'key': '14', 'name': 'Comedy Specials'},
    {   'id': 'xxxx', 'key': '2', 'name': 'Movies'},
    {   'id': 'xxxx', 'key': '12', 'name': 'Movies Anime'},
    {   'id': 'xxxx', 'key': '4', 'name': 'Music'},
    {   'id': 'xxxx', 'key': '3', 'name': 'TV Shows'},
    {   'id': 'xxxx', 'key': '11', 'name': 'TV Shows Anime'},
    {   'id': 'xxxx', 'key': '10', 'name': 'Workout'}]

```

looks like I might be a pull behind.

### DIFF
--- a/plexapi/myplex.py
+++ b/plexapi/myplex.py
@@ -136,20 +136,15 @@ class MyPlexAccount(PlexObject):
         data = self.query(MyPlexDevice.key)
         return [MyPlexDevice(self, elem) for elem in data]
 
-    def _headers(self, **kwargs):
-        """ Returns dict containing base headers for all requests to the server. """
-        headers = BASE_HEADERS.copy()
-        if self._token:
-            headers['X-Plex-Token'] = self._token
-        headers.update(kwargs)
-        return headers
-
     def query(self, url, method=None, headers=None, timeout=None, **kwargs):
         method = method or self._session.get
+        delim = '&' if '?' in url else '?'
+        url = '%s%sX-Plex-Token=%s' % (url, delim, self._token)
         timeout = timeout or TIMEOUT
         log.debug('%s %s %s', method.__name__.upper(), url, kwargs.get('json', ''))
-        headers = self._headers(**headers or {})
-        response = method(url, headers=headers, timeout=timeout, **kwargs)
+        allheaders = BASE_HEADERS.copy()
+        allheaders.update(headers or {})
+        response = method(url, headers=allheaders, timeout=timeout, **kwargs)
         if response.status_code not in (200, 201, 204):
             codename = codes.get(response.status_code)[0]
             errtext = response.text.replace('\n', ' ')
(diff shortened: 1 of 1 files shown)


pkkid:
This change looks good to me. @Hellowlol was working in that area more recently, so would like his input as well before we merge.

Hellowlol:
So this pr isn't needed it should be supported atm but isn't since as i apparently haven't tested this good enough. Add is section as a last minute change. I wasnt aware that a Element was False...

Also the naming could probably be better. 

to get a users shared section you can do:
````
account = MyPlexAccount()
user = account.user('Hellowlol')
# servers is a list since you can share more servers
shared_sections = users.servers[0].sections()
````

**Comment by blacktwin** [comment]
Using your method:
```python
account = MyPlexAccount()
user = account.user('Hellowlol')
# servers is a list since you can share more servers
shared_sections = users.servers[0].sections()
```
Doesn't produce the desired output but you're saying the it's not yet supported?

Hellowlol:
It was intented to be supported but wasnt since sections would always return a empty list. Now it returns a list of section objects where you have access to the same info as in your list of dicts and some more.

Hellowlol:
You need to pull from master to use this feature btw.

Hellowlol:
I think im gonna rename the sections as it doesnt describe it will some might think its the same library sections

blacktwin:
Trying:
```python
user = account.user('username')
shared_sections = user.servers[0].sections()
print(shared_sections)
```
Produces:
```python
File "C:\Python27\lib\site-packages\plexapi\myplex.py", line 508, in sections
    if isinstance(section, ElementTree.Element):
TypeError: isinstance() arg 2 must be a class, type, or tuple of classes and types
```
Am I missing something? I've updated plexapi just before running.

Hellowlol:
I'm sure I tested this before I merged. Can you try with python 3?

Hellowlol:
Anyway I'll fix this later today. If you figure this you can send a pr

Hellowlol:
It was a python2 bug. Its fixed now.

blacktwin:
Wow! Very NICE! Thanks!

PR #224, opened 2017-10-25
Covcov

Hellowlol:


plexapi/client.py
```diff
@@ -444,7 +444,7 @@ def playMedia(self, media, offset=0, **params):
         if not self.product != 'OpenPHT':
             try:
                 self.sendCommand('timeline/subscribe', port=server_url[1].strip('/'), protocol='http')
-            except:
+            except:  # noqa: E722
```
pkkid: if you say `except Exception:` I believe pyflakes wont yell at you.

Hellowlol:
Btw, is the plexapi username password and token in the test computers env? I would like to swap to using env as the current solution uses the config. And i got my own server in the config.. It make local testing painfull.

tests/test_video.py
```diff
@@ -61,6 +60,15 @@ def test_video_Movie_isPartialObject(movie):
     assert movie.isPartialObject()
 
 
+def test_video_Movie_delete_part(movie, mocker):
+    # we need to reload this as there is a bug in part.delete
+    # See https://github.com/pkkid/python-plexapi/issues/201
+    m = movie.reload()
+    for part in m.iterParts():
```
Hellowlol: We should iterate over media

Hellowlol:
I'll go ahead and merge this. It's gonna be a while until I have time to complete the rest

pkkid:
:+1:

PR #233, opened 2017-11-17
deleteFromUser()

blacktwin:
Allow server admin to delete playlists that were copiedToUser.

### DIFF
--- a/plexapi/playlist.py
+++ b/plexapi/playlist.py
@@ -132,3 +132,14 @@ class Playlist(PlexPartialObject, Playable):
         # Login to your server using your friends credentials.
         user_server = PlexServer(self._server._baseurl, token)
         return self.create(user_server, self.title, self.items())
+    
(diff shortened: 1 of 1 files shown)


Hellowlol:
I guess there are several cases where you want to login as a user. What about adding a loginAs on plex server.

Hellowlol:
Thanks for your contribution.  I havn't forgotten about this pr, ill get to this in a couple of weeks.

blacktwin:
I don't think this is needed any more. Using an user's token to create a user server object will allow for deleting playlist for the user. example:

```python
plex = PlexServer(PLEX_URL, PLEX_TOKEN, session=sess)
user_acct = account.user('USERNAME')
user_plex = PlexServer(PLEX_URL, user_acct.get_token(plex.machineIdentifier))
userplex.playlist('Test Playlist').delete
```

PR #245, opened 2018-01-11
update plexurl

ninthwalker:
It appears plex is doing away with the plexapp.com url.
It no longer has a valid SAN cert for that domain as it redirects to plex.tv now.
Update myplex.py to reflect new signin URL.


Hellowlol:
Thanks btw :)

PR #247, opened 2018-01-11
my.plexapp.com no longer valid fix

ryanm101:


ryanm101:
Someone beat me to it
https://github.com/pkkid/python-plexapi/pull/245

PR #252, opened 2018-02-18
Add `updateProgress` function to `Video`

d1vanloon:
Allows the client to update the playback position for videos.

### DIFF
--- a/plexapi/video.py
+++ b/plexapi/video.py
@@ -77,6 +77,21 @@ class Video(PlexPartialObject):
         self._server.query(key)
         self.reload()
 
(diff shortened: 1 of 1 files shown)


Hellowlol:
Can you document all the different state that's available. This should be added to audio aswell so we can set progress on a song.

d1vanloon:
I looked for documentation on the `state` parameter, and I couldn't find any. I don't believe it's needed. After reading your comment, I considered changing the name of the function to relate to "view offset," since that is the value that is updated (at least on videos). What are your thoughts?

(I see you've gone ahead and merged this and added a function for audio - I apologize for my delay in responding)

Issue #259, opened 2018-03-05
[Music library] add userRating


patzm:
I found that the _ratingCount_ property of [plexapi.audio.Track](http://python-plexapi.readthedocs.io/en/latest/modules/audio.html#plexapi.audio.Track) does not hold the rating a user assigned to the track. It contains seemingly random, positive integer values.

The correct rating is however available in the __data.attrib['userRating']_ field of the object. The key 'userRating' is non-existent if the track is not rated, yet.

I propose to expose this attribute as a public variable and fix the documentation (see link above). What are your thoughts on this?

Hellowlol:
I doubt plex will inject random stuff as we need to fixure out where that value comes from and documentet it. We should have userRating. Can you send a pr for it?

MattMcDowall:
> we need to fixure out where that value comes from and documentet it.

The `ratingCount` property is a reflection of the listener counts on [Last.fm](https://last.fm). I suspected this for a while, but to confirm I looked at the numbers for one of my relatively obscure albums (_Geared & Primed_ by The Royal Court of China). Sure enough, the counts match exactly.

I get:
```
84	 1. Geared and Primed
113	 2. Half the Truth
63	 3. It Came Crashing down the Staircase
66	 4. Six Empty Bottles
4	 5. Mr. Indecision
4	 6. Tijuana Go!
57	 7. Dragon Park
61	 8. So, Yer Love Is True
61	 9. This Time Around
52	10. Take Me Down
```

And Last.fm shows:
![image](https://user-images.githubusercontent.com/51373511/171498298-1f11ce5c-668c-4129-ac7c-5f5bea207d7f.png)

If you've used Last.fm, you know it's susceptible to inconsistencies in artist names, track names, etc. This is why "Mr. Indecision" & "Tijuana Go!" have anomalously low ratings here.

But that's clearly where Plex is getting its ratingCount number from.

patzm:
Cool! So you could actually update the docs now with this information right? Maybe PR it?

MattMcDowall:
Honestly, I would love to, but I don't know how. I'm really just learning GitHub, and I haven't contributed to a project like this before. I'm familiar with the process of a PR, in theory, at least. But can you point me to where I can make a change to the docs?

patzm:
fork this repo
![image](https://user-images.githubusercontent.com/5219726/171658343-c95c135c-efb2-4dfd-8ea8-81beff6625ff.png)

[...]

PR #263, opened 2018-03-21
added userRating property to Track

patzm:
I figured that using `float `is better than `int`, even though it is an integer right now. This should make it ready for any changes in the future if Plex starts implementing continuous ratings.

I still couldn't figure out what the `ratingCount` property is supposed to express. I ran an analysis on many files in my server. Approx. half of them do not have a value assigned to this property. The other half seems to have random positive values that do not correlate with the (start) rating they have. Also it is not related to the play count.

### DIFF
--- a/plexapi/audio.py
+++ b/plexapi/audio.py
@@ -251,7 +251,8 @@ class Track(Audio, Playable):
             parentThumb (str): URL to album thumbnail image.
             parentTitle (str): Name of the album for this track.
             primaryExtraKey (str): Unknown
-            ratingCount (int): Rating of this track (1-10?)
+            ratingCount (int): Unknown
+            userRating (float): Rating of this track (0.0 - 10.0) equaling (0 stars - 5 stars)
             viewOffset (int): Unknown
             year (int): Year this track was released.
             sessionKey (int): Session Key (active sessions only).
@@ -284,6 +285,7 @@ class Track(Audio, Playable):
         self.parentTitle = data.attrib.get('parentTitle')
         self.primaryExtraKey = data.attrib.get('primaryExtraKey')
         self.ratingCount = utils.cast(int, data.attrib.get('ratingCount'))
+        self.userRating = utils.cast(float, data.attrib.get('userRating', 0))
         self.viewOffset = utils.cast(int, data.attrib.get('viewOffset', 0))
         self.year = utils.cast(int, data.attrib.get('year'))
         self.media = self.findItems(data, media.Media)


patzm:
referecing #259

odinian:
Sorry for my ignorance, but how can I get this branch? I would like to be able to read and write the userRating.

patzm:
either you hop over to [my repository](https://github.com/patzm/python-plexapi/tree/pr/259) or access it via a protected variable for now:
`track._data.attrib['userRating']`

odinian:
thanks. I think I just realized the API is only for reading. Are there any "set" commends to write data back to Plex?

Hellowlol:
Not even close. I suggest you start by reading the docs. http://python-plexapi.readthedocs.io/en/latest/ You can also see some examples in the tests.

odinian:
Thanks. I spent quite a bit of time reading those. Very hard to understand with so few examples. Still, I was able to extra all the data I wanted (except userRating). I really wasn't looking for "set" until after I got the "get" working. Oh well. Thanks again.

Hellowlol:
I think your looking for object.edit

patzm:
I have to agree to some degree with @odinian regarding the documentation. I was also looking for a "setting" functionality and totally overlooked the `object.edit` methods. For sure that was my fault, but maybe the fact that several people overlooked it hints towards that the docu is not explicit enough on this aspect.
However the above comments are not related to this issue.

odinian:
thanks for all your help. @patzm , will object.edit work for userRating in your branch?

patzm:
yes it does. See the following example:
`track.edit(**{'userRating.value': 10})`
Is this the way it is supposed to be?
thanks to [Mattie432](https://github.com/patzm/plex-music-rating-sync/issues/1) for the hint

MattMcDowall:
> I still couldn't figure out what the `ratingCount` property is supposed to express.

The `ratingCount` property is a reflection of the listener counts on [Last.fm](https://last.fm). I suspected this for a while, but to confirm I looked at the numbers for one of my relatively obscure albums (_Geared & Primed_ by The Royal Court of China). Sure enough, the counts match exactly.

I get:
```
84	 1. Geared and Primed
113	 2. Half the Truth
63	 3. It Came Crashing down the Staircase
66	 4. Six Empty Bottles
4	 5. Mr. Indecision
4	 6. Tijuana Go!
57	 7. Dragon Park
61	 8. So, Yer Love Is True
61	 9. This Time Around
52	10. Take Me Down
```

And Last.fm shows:
image

If you've used Last.fm, you know it's susceptible to inconsistencies in artist names, track names, etc. This is why "Mr. Indecision" & "Tijuana Go!" have anomalously low ratings here.

But that's clearly where Plex is getting its ratingCount number from.

PR #270, opened 2018-05-12
Function update timeline for base.py

tijder:
Function to update the timeline for a payable item. With this the item will be showed in the now playing screen.

### DIFF
--- a/plexapi/base.py
+++ b/plexapi/base.py
@@ -574,9 +574,11 @@ class Playable(object):
                 state (string): state of the video, default 'stopped'
                 duration (int): duration of the item
         """
-        durationStr = ''
+        durationStr = '&duration='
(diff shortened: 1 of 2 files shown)


pkkid:
Thanks! -- If you have a few extra minutes, can you add a test for this?  If not, that's fine, I'll add a test when I have time before merging this in.

tijder:
I wrote a basic test. I only not familiar with python tests. I don't know how to write a test for testing if the timeline is added to the now playing screen.

Hellowlol:
Can we just set the duration automatically if duration is omitted?
````
if duration is None:
    duration = self.duration
....
````

tijder:
Good idea. I added it in the code.

Issue #272, opened 2018-05-29
TypeError: download() missing 1 required positional argument: 'token'


pkkid:
```python
Traceback (most recent call last):
  File "./plex-download.py", line 88, in <module>
    session=item._server._session, showstatus=True)
TypeError: download() missing 1 required positional argument: 'token'
```

Hellowlol:
This was fixed awhile back.

PR #278, opened 2018-07-13
Fix issue #272 (No authToken) and adding getToken program

MichelfrancisBustillos:
Fix Issue with plex-download.py by adding authToken.

Add plex-gettoken.py as a simple way to get authToken. Useable as standalone or as part of a larger program with slight modification.

### DIFF
--- a/tools/plex-download.py
+++ b/tools/plex-download.py
@@ -63,8 +63,7 @@ def get_item_from_url(url):
         raise SystemExit('Unknown or ambiguous client id: %s' % clientid)
     server = servers[0].connect()
     return server.fetchItem(key)
-
-
+	
 if __name__ == '__main__':
     # Command line parser
     from plexapi import CONFIG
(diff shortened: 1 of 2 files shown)


mikes-nasuni:
Does plex-gettoken.py do the same thing as plex-listtokens.py?

MichelfrancisBustillos:
It does not. plex-listtokens.py lists all existing tokens for all clients. plex-gettoken.py pulls a new token that can be used to authenticate. It's mostly intended to get a token for using in conf.py or for modification and integration in future scripts that utilize the API. I wrote it up because I had difficulty getting a simple, working authToken via the CLI, primarily to solve issue #272 with plex-download.py. I have since been shown a much more simple method to solve that issue (see above) but decided to push the method I had created anyway in case it is of use to others.

mikes-nasuni:
Cool, thanks for the fix and the explanation.

PR #282, opened 2018-08-19
[WIP] Sync support

### LINKED ISSUE #134: Support Sync
This has never worked, lets make it work.

andrey-yantsen:
Hi there! I've started working on Mobile Sync support. If anybody has any comments on the code, or just any feature requests — feel free to tell :)

At the beginning the code can do few useful things:

1. Initialise the "sync" env: tell the plex.tv that the client supports syncing, and the client can be used as sync-target
2. Unfortunately I have to "fake" used platform/device info, because transcoding is depends on this values
3. And finally you can download converted files, by iterating over media parts and looking for ones with required `syncItemId`

The reason why I'm started this is simple: I have a WD My Passport Pro which has the Plex onboard and I'd like to have my "travel" library in sync with the main one. I've already started the project for this (andrey-yantsen/plexiglas) but for now there is nothing to see there.

fix #134

### DIFF
--- a/plexapi/myplex.py
+++ b/plexapi/myplex.py
@@ -3,7 +3,7 @@ import copy
 import requests
 import time
 from requests.status_codes import _codes as codes
-from plexapi import BASE_HEADERS, CONFIG, TIMEOUT
+from plexapi import BASE_HEADERS, CONFIG, TIMEOUT, X_PLEX_IDENTIFIER
 from plexapi import log, logfilter, utils
 from plexapi.base import PlexObject
 from plexapi.exceptions import BadRequest, NotFound
@@ -11,6 +11,7 @@ from plexapi.client import PlexClient
 from plexapi.compat import ElementTree
 from plexapi.library import LibrarySection
 from plexapi.server import PlexServer
+from plexapi.sync import SyncList
 from plexapi.utils import joinArgs
 
 
@@ -377,6 +378,14 @@ class MyPlexAccount(PlexObject):
             params['optOutLibraryStats'] = int(library)
         url = 'https://plex.tv/api/v2/user/privacy'
         return self.query(url, method=self._session.put, params=params)
+
+    def syncItems(self, clientId=None):
+        if clientId is None:
+            clientId = X_PLEX_IDENTIFIER
+
+        data = self.query(SyncList.key.format(clientId=clientId))
+
+        return SyncList(self, data)
 
 
 class MyPlexUser(PlexObject):
--- a/plexapi/sync.py
+++ b/plexapi/sync.py
@@ -1,31 +1,53 @@
 # -*- coding: utf-8 -*-
 import requests
-from plexapi import utils
+import plexapi
 from plexapi.exceptions import NotFound
+from plexapi.base import PlexObject
 
 
-class SyncItem(object):
-    """ Sync Item. This doesn't current work. """
-    def __init__(self, device, data, servers=None):
-        self._device = device
-        self._servers = servers
-        self._loadData(data)
+def init(replace_provides=False):
+    if replace_provides or not plexapi.X_PLEX_PROVIDES:
+        plexapi.X_PLEX_PROVIDES = 'sync-target'
+    else:
+        plexapi.X_PLEX_PROVIDES += ',sync-target'
+
+    plexapi.BASE_HEADERS['X-Plex-Sync-Version'] = '2'
+    plexapi.BASE_HEADERS['X-Plex-Provides'] = plexapi.X_PLEX_PROVIDES
+
+    # mimic iPhone SE
+    plexapi.X_PLEX_PLATFORM = 'iOS'
+    plexapi.X_PLEX_PLATFORM_VERSION = '11.4.1'
+    plexapi.X_PLEX_DEVICE = 'iPhone'
+
+    plexapi.BASE_HEADERS['X-Plex-Platform'] = plexapi.X_PLEX_PLATFORM
+    plexapi.BASE_HEADERS['X-Plex-Platform-Version'] = plexapi.X_PLEX_PLATFORM_VERSION
+    plexapi.BASE_HEADERS['X-Plex-Device'] = plexapi.X_PLEX_DEVICE
+    plexapi.BASE_HEADERS['X-Plex-Model'] = '8,4'
+    plexapi.BASE_HEADERS['X-Plex-Vendor'] = 'Apple'
+
+
+class SyncItem(PlexObject):
+    TAG = 'SyncItem'
+
+    def __init__(self, server, data, initpath=None, clientIdentifier=None):
+        super().__init__(server, data, initpath)
+        self.clientIdentifier = clientIdentifier
 
     def _loadData(self, data):
         self._data = data
-        self.id = utils.cast(int, data.attrib.get('id'))
-        self.version = utils.cast(int, data.attrib.get('version'))
+        self.id = plexapi.utils.cast(int, data.attrib.get('id'))
+        self.version = plexapi.utils.cast(int, data.attrib.get('version'))
         self.rootTitle = data.attrib.get('rootTitle')
         self.title = data.attrib.get('title')
         self.metadataType = data.attrib.get('metadataType')
         self.machineIdentifier = data.find('Server').get('machineIdentifier')
-        self.status = data.find('Status').attrib.copy()
+        self.status = Status(self._server, data.find('Status'))
         self.MediaSettings = data.find('MediaSettings').attrib.copy()
         self.policy = data.find('Policy').attrib.copy()
         self.location = data.find('Location').attrib.copy()
 
     def server(self):
-        server = list(filter(lambda x: x.machineIdentifier == self.machineIdentifier, self._servers))
+        server = list(filter(lambda x: x.clientIdentifier == self.machineIdentifier, self._server.resources()))
         if 0 == len(server):
             raise NotFound('Unable to find server with uuid %s' % self.machineIdentifier)
         return server[0]
@@ -35,8 +57,37 @@ class SyncItem(object):
         key = '/sync/items/%s' % self.id
         return server.fetchItems(key)
 
-    def markAsDone(self, sync_id):
+    def markAsDone(self):
         server = self.server().connect()
         url = '/sync/%s/%s/files/%s/downloaded' % (
-            self._device.clientIdentifier, server.machineIdentifier, sync_id)
+            self.clientIdentifier, server.machineIdentifier, self.id)
         server.query(url, method=requests.put)
+
+
+class SyncList(PlexObject):
+    key = 'https://plex.tv/devices/{clientId}/sync_items'
+    TAG = 'SyncList'
+
+    def _loadData(self, data):
+        self._data = data
+        self.clientId = data.attrib.get('clientIdentifier')
+
+        for elem in data:
+            if elem.tag == 'SyncItems':
+                self.items = self.findItems(elem, SyncItem)
+
+
+class Status(PlexObject):
+    TAG = 'Status'
+
+    def _loadData(self, data):
+        self._data = data
+        self.failureCode = data.attrib.get('failureCode')
+        self.failure = data.attrib.get('failure')
+        self.state = data.attrib.get('complete')
+        self.itemsCount = plexapi.utils.cast(int, data.attrib.get('itemsCount'))
+        self.itemsCompleteCount = plexapi.utils.cast(int, data.attrib.get('itemsCompleteCount'))
+        self.totalSize = plexapi.utils.cast(int, data.attrib.get('totalSize'))
+        self.itemsDownloadedCount = plexapi.utils.cast(int, data.attrib.get('itemsDownloadedCount'))
+        self.itemsReadyCount = plexapi.utils.cast(int, data.attrib.get('itemsReadyCount'))
+        self.itemsSuccessfulCount = plexapi.utils.cast(int, data.attrib.get('itemsSuccessfulCount'))
--- a/plexapi/media.py
+++ b/plexapi/media.py
@@ -92,6 +92,11 @@ class MediaPart(PlexObject):
         self.indexes = data.attrib.get('indexes')
         self.key = data.attrib.get('key')
         self.size = cast(int, data.attrib.get('size'))
+        self.decision = data.attrib.get('decision')
+        self.optimizedForStreaming = cast(bool, data.attrib.get('optimizedForStreaming'))
+        self.syncItemId = cast(int, data.attrib.get('syncItemId'))
+        self.syncState = data.attrib.get('syncState')
+        self.videoProfile = data.attrib.get('videoProfile')
         self.streams = self._buildStreams(data)
 
     def _buildStreams(self, data):


Hellowlol:
Hi! Thanks for the PR!

How do you actually sync the items? Can you add a test?

I would expect that there was a sync function inside the client so you could just pass a list of objects and that would be it.

Any info needed for the transcode can be picked up there this way we don’t need to mess with the init method.

andrey-yantsen:
Just to clarify: the PR is not finished now, it's an early alpha version :)

> How do you actually sync the items?

The list of "items-to-sync" is controlled in Plex Web, you can add any item to the list for a specific device.

In the following code block I will show the way to download all suitable content (actually I've not yet reached the "downloading" stage in my project, so I'm not sure if this really working now):

```
from plexapi import sync, utils
from plexapi.myplex import MyPlexAccount
sync.init()
plex = MyPlexAccount()

for item in plex.syncItems().items:
    if item.status.state != 'complete':
        continue

    for media in item.getMedia():
        for part in media.iterParts():
            filename = '%s.%s' % (item._prettyfilename(), part.container)
            url = item._server.url(part.key)
            filepath = utils.download(url, token=plex.authenticationToken, filename=filename, savepath=os.getcwd(),
                                      session=item.server()._session, showstatus=True)

```

> Can you add a test?

Yep, of course, on some latter stage :)

> I would expect that there was a sync function inside the client so you could just pass a list of objects and that would be it.
> Any info needed for the transcode can be picked up there this way we don’t need to mess with the init method.

As far as I can see the only way for Plex to detect transcoding options, is thru target platform. Here is what I see in the Plex logs with real platform info:

```
ERROR - Unable to find client profile for device; platform=Darwin, platformVersion=17.7.0, device=
Darwin, model=
```
And the process goes smoothly when I mimic my iPhone SE.

Also here is the request which should be sent to add an item to sync list:

```
POST https://plex.tv/devices/779755E8-54E2-4D0F-A7D9-E2FF8171CC55/sync_items?X-Plex-Device-Name=Yantsen%20-%20iPhone%20An HTTP/1.1

Content-Type:	application/x-www-form-urlencoded

SyncItem[Location][uri]:        library://ac0a6a57-522b-4dcf-9cb8-0a58b16c6726/item/%2Flibrary%2Fmetadata%2F2428
SyncItem[MediaSettings][audioBoost]:100
SyncItem[MediaSettings][maxVideoBitrate]:4000
SyncItem[MediaSettings][musicBitrate]:192
SyncItem[MediaSettings][photoQuality]:74
SyncItem[MediaSettings][photoResolution]:1920x1080
SyncItem[MediaSettings][subtitleSize]:100
SyncItem[MediaSettings][videoQuality]:100
SyncItem[MediaSettings][videoResolution]:1280x720
SyncItem[Policy][scope]:        all
SyncItem[Policy][unwatched]:    0
SyncItem[Policy][value]:        0
SyncItem[contentType]:          video
SyncItem[machineIdentifier]:    607c8c938b50eef734456f8b9da94b5d02339ce5
SyncItem[metadataType]:         movie
SyncItem[rootTitle]:            Test
SyncItem[title]:                Test1
```

As you can see there are nothing here about container and/or supported codecs list, only basic/common transcoding stuff.

Hellowlol:
You dont really need to do anything to make the sync happens. Seems you just upload zhe shit to a queue, and the client handles the rest from there.

Anyway to send a sync command should be as easy as:

inside client.py
````

def sync_item(self, item): 
    #magic happens here.
    # Maybe grab stuff that you need from the client like platform etc
    # name, title from item etc.

   
````

Looking forward to test the pr :)

**Comment by andrey-yantsen** [comment]
> I dont think syncing to computers is supported by plex at all.

It's not supported out-of-the-box, but we're the engineers, and we can force software to do what we want! :)

> Do you just want to copy things from another shared plex server to your own?

I want to copy optimised copies of things from any plex server to an external hdd

**Comment by Hellowlol** [comment]
Oh, thats easy.  Its pretty much supported out of the box.

See the docs for.

https://python-plexapi.readthedocs.io/en/latest/modules/video.html?highlight=getStreamURL%20#plexapi.video.Movie.download

note the comment about **kwargs – Additional options passed into getStreamURL().
You can use this to transcode.

**Comment by andrey-yantsen** [comment]
Yep, I saw this option, I plan to support it in my project, but it's not on my priority list...

`getSteamURL()` not that powerful, easy and cute like Mobile Sync: e.g. with Sync you can mark to sync "3 unwatched episodes", "3 unwatched movies with highest rating", "3 unwatched movies released recently", and so on. And all those media would be automatically rotated while you marking something as watched, when new releases become available.

**Comment by Hellowlol** [comment]
I think im missing something here. Do you expect to do this from the plex ui?
If would be easy to get movies/episodes based on some condition run them in some cron script.

**Comment by andrey-yantsen** [comment]
> Do you expect to do this from the plex ui?
You're exactly right :) https://support.plex.tv/articles/201082477-quick-guide-to-mobile-sync/

> If would be easy to get movies/episodes based on some condition run them in some cron script.
Man... I just like the Mobile Sync approach, where you don't have to code the filters by yourself, but you can easily choose the required things with UI and this UI doesn't require any dev-knowledge at all, even my wife would be able to use it! :D

**Comment by Hellowlol** [comment]
Oh, i think i get it now. You faking so plexapi is detected as a client in the plex webui... imo that should be done with your own project!

Nice idea btw!

File: plexapi/base.py [hunk 1]
```diff
@@ -144,7 +144,8 @@ def fetchItems(self, ekey, cls=None, **kwargs):
             and attrs. See :func:`~plexapi.base.PlexObject.fetchItem` for more details
             on how this is used.
         """
-        data = self._server.query(ekey)
+        timeout = kwargs.pop('timeout', None)
```
mikes-nasuni: This should probably be added as a real keyword arg timeout=None in the function definition and not pulled out of kwargs.

plexapi/myplex.py
```diff
@@ -707,7 +707,7 @@ def delete(self):
         self._server.query(key, self._server._session.delete)
 
 
-def _connect(cls, url, token, timeout, results, i):
+def _connect(cls, url, token, timeout, fast, results, i, connected_event=None):
```
mikes-nasuni: If X_PLEX_ENABLE_FAST_CONNECT is a global configuration option. Should we instead always use fast=True or false based on that, rather than passing it in as an argument?

plexapi/utils.py
```diff
@@ -152,15 +152,18 @@ def threaded(callback, listargs):
...
         args += [results, len(results)]
         results.append(None)
-        threads.append(Thread(target=callback, args=args))
+        threads.append(threading.Thread(target=callback, args=args, kwargs=dict(connected_event=connected_event)))
         threads[-1].setDaemon(True)
         threads[-1].start()
-    for thread in threads:
-        thread.join()
-    return results
+    while not connected_event.is_set():
```
mikes-nasuni: Just curious, what was the issue that led to this change?

andrey-yantsen: Most times I run the my things outside my home network, when the local IP of the server is not available, and it takes annoyingly long time to establish the connection

andrey-yantsen:
@mikes-nasuni thank you for the fast reaction, I've made the changes which you've suggested :)

plexapi/utils.py
```diff
@@ -152,15 +152,18 @@ def threaded(callback, listargs):
...
+        if all([not t.is_alive() for t in threads]):
+            break
+        time.sleep(0.05)
+    return list(filter(lambda r: r is not None, results))
```
Hellowlol: Use a list comprehension

plexapi/sync.py
```diff
@@ -1,42 +1,103 @@
...
         key = '/sync/items/%s' % self.id
-        return server.fetchItems(key)
+        return server.fetchItems(key, timeout=timeout)
 
-    def markAsDone(self, sync_id):
-        server = self.server().connect()
-        url = '/sync/%s/%s/files/%s/downloaded' % (
-            self._device.clientIdentifier, server.machineIdentifier, sync_id)
-        server.query(url, method=requests.put)
+    def markDownloaded(self, media: Playable):
```
Hellowlol: This need to be support python 2.

plexapi/sync.py
```diff
@@ -1,42 +1,103 @@
...
-        self.status = data.find('Status').attrib.copy()
+        self.status = Status(self._server, data.find('Status'))
         self.MediaSettings = data.find('MediaSettings').attrib.copy()
         self.policy = data.find('Policy').attrib.copy()
         self.location = data.find('Location').attrib.copy()
 
     def server(self):
-        server = list(filter(lambda x: x.machineIdentifier == self.machineIdentifier, self._servers))
+        server = list(filter(lambda x: x.clientIdentifier == self.machineIdentifier, self._server.resources()))
         if 0 == len(server):
```
Hellowlol: List comp please

plexapi/sync.py
```diff
@@ -1,42 +1,103 @@
...
             raise NotFound('Unable to find server with uuid %s' % self.machineIdentifier)
         return server[0]
 
-    def getMedia(self):
+    def getMedia(self, timeout=None):
```
Hellowlol: What up with the timeout? Why not just change the plex.Timeout?

andrey-yantsen: Not all operations require 60s+ timeout, and wrapping `getMedia()` calls with changing timeout back-and-forth would look strange.
You think that just globally change `TIMEOUT` in my code would be more than enough and wouldn't cause any strange problems?

andrey-yantsen:
@Hellowlol thank you for the review! I've fixed most of the issues, but I'm not sure on `timeout` arg in `getMedia()` / `fetchItems()`.

plexapi/utils.py
mikes-nasuni: Makes sense, thanks.

plexapi/sync.py
mikes-nasuni: > You think that just globally change TIMEOUT in my code would be more than enough and wouldn't cause any strange problems?

It shouldn't cause any problems. timeout is essentially just telling the requests module when to give up if not getting a response. It's really only useful to change to poor or slow network issues. Having different timeouts for different calls in to the API seems incorrect unless you have a really strong case for it.

plexapi/myplex.py
```diff
@@ -698,7 +707,7 @@ def delete(self):
         self._server.query(key, self._server._session.delete)
 
 
-def _connect(cls, url, token, timeout, results, i):
+def _connect(cls, url, token, timeout, results, i, connected_event=None):
```
mikes-nasuni: What is connected_event used for, can you add a comment in the doc string here about what what is is?

plexapi/__init__.py
```diff
@@ -17,6 +17,7 @@
 VERSION = '3.0.6'
 TIMEOUT = CONFIG.get('plexapi.timeout', 30, int)
 X_PLEX_CONTAINER_SIZE = CONFIG.get('plexapi.container_size', 100, int)
+X_PLEX_ENABLE_FAST_CONNECT = CONFIG.get('plexapi.enable_fast_connect', False, bool)
```
mikes-nasuni: Can you please update the configuration.rst file with what X_PLEX_ENABLE_FAST_CONNECT is and when we would want to use it.

https://github.com/pkkid/python-plexapi/blob/master/docs/configuration.rst

plexapi/sync.py
```diff
@@ -1,31 +1,58 @@
...
-        self._loadData(data)
+class SyncItem(PlexObject):
+    TAG = 'SyncItem'
+
+    def __init__(self, server, data, initpath=None, clientIdentifier=None):
+        super(SyncItem, self).__init__(server, data, initpath)
+        self.clientIdentifier = clientIdentifier
 
     def _loadData(self, data):
         self._data = data
```
mikes-nasuni: There should be a documentation header on head class in this file explaining all the attributes and why they exist. A good example format is the video.py file.

https://github.com/pkkid/python-plexapi/blob/master/plexapi/video.py

plexapi/sync.py
```diff
@@ -35,8 +62,53 @@ def getMedia(self):
...
+        Mark the file as downloaded within current SyncItem
+
+        :param media:
+        :type media: base.Playable
+        """
+        url = '/sync/%s/item/%s/downloaded' % (self.clientIdentifier, media.ratingKey)
+        media._server.query(url, method=requests.put)
+
+
+class SyncList(PlexObject):
```
mikes-nasuni: Also needs documentation.

mikes-nasuni:
This is starting to shape up nicely, thanks.

plexapi/config.py
```diff
@@ -60,4 +60,5 @@ def reset_base_headers():
         'X-Plex-Device': plexapi.X_PLEX_DEVICE,
         'X-Plex-Device-Name': plexapi.X_PLEX_DEVICE_NAME,
         'X-Plex-Client-Identifier': plexapi.X_PLEX_IDENTIFIER,
+        'X-Plex-Sync-Version': '2',
```
andrey-yantsen: I've added this because it doesn't require anything from the client, it just reports to the server the we can work with Sync. Basically without this header you can't add new SyncItems.

andrey-yantsen:
Hey guys, please review :) Looks like I almost finished with the development (actually I already have my app syncing videos for me for almost a week now), the one huge thing which is left — is tests... And make all of this work with audio/photos, but it shouldn't be that hard at current stage. I hope I will be able to finish them during next week.

Hellowlol:
Sorry, I’ll review this later today. Been busy.

andrey-yantsen:
FYI there is a typo in `__init__.py`:

```
X_PLEX_PLATFORM = CONFIG.get('header.platorm', uname()[0])
```

`platorm` instead of `platform`. Probably we should fix this, but this would be a breaking change :(

andrey-yantsen:
Ok. At the moment I have tests for video/show/season/episode adding to synclist and reading from it. Here is what left to do:

* Test downloading
* Add method sync() for audio, photo & playlist

Anything else?

plexapi/library.py
```diff
@@ -543,6 +543,53 @@ def _cleanSearchSort(self, sort):
             raise BadRequest('Unknown sort dir: %s' % sdir)
         return '%s:%s' % (lookup[scol], sdir)
 
+    def sync(self, client, policy, media_settings, title=None, sort=None, libtype=None, **kwargs):
+        """ Add current library section as sync item for specified device.
+            See description of :func:`plexapi.library.LibraySection.search()` for details about filtering / sorting.
+
```
Hellowlol: Example is good but can you add docs string for args?

andrey-yantsen: Yep, sure, I'm working on tests now, and fixing all the docs :)

plexapi/myplex.py
```diff
@@ -378,6 +379,57 @@ def optOut(self, playback=None, library=None):
         url = 'https://plex.tv/api/v2/user/privacy'
         return self.query(url, method=self._session.put, params=params)
 
+    def syncItems(self, clientId=None):
+        """ Returns an instance of :class:`~plexapi.sync.SyncItems` for specified client
+
+            Arguments:
+                 clientId (str): an identifier of a client for which you need to get SyncItems. Would be set to current
```
Hellowlol: allow for both a client and a clientid as a string.

plexapi/myplex.py
```diff
@@ -378,6 +379,57 @@ def optOut(self, playback=None, library=None):
...
+
+        data = self.query(SyncList.key.format(clientId=clientId))
+
+        return SyncList(self, data)
+
+    def sync(self, client, sync_item):
+        """ Adds specified sync item for the client
+
+            Arguments:
+                 client (:class:`~plexapi.myplex.MyPlexDevice`): pass
```
Hellowlol: doc what this client, sync_item is used for.

plexapi/server.py
```diff
@@ -422,6 +422,16 @@ def url(self, key, includeToken=None):
             return '%s%s%sX-Plex-Token=%s' % (self._baseurl, key, delim, self._token)
         return '%s%s' % (self._baseurl, key)
 
+    def refreshSynclist(self):
+        return self.query('/sync/refreshSynclists', self._session.put)
```
Hellowlol: add missing doc string

plexapi/server.py
```diff
@@ -422,6 +422,16 @@ def url(self, key, includeToken=None):
...
+
+    def refreshContent(self):
+        return self.query('/sync/refreshContent', self._session.put)
```
Hellowlol: add missing doc string

plexapi/sync.py
```diff
@@ -1,42 +1,256 @@
...
+            * when you create an item for a season 3 of show Example, the value would be `Season 3`
+            * when you set to sync all your movies in library named "My Movies" to value would be `My Movies`
+
+        title (str): the title which you've set when created the sync item
+        metadataType (str): the type of media which hides inside, can be `episode`, `movie`, etc.
+        contentType (str): basic type of the content: `video` or `audio`
+        status (:class:`~plexapi.sync.Status`): current status of the sync
+        mediaSettings (:class:`~plexapi.sync.MediaSettings`): media transcoding settings used for the item
+        policy (:class:`~plexapi.sync.Policy`): the policy of which media to sync
+        location (str): unknown
```
Hellowlol: self.clientidentifier is not listed as a attribute.

plexapi/sync.py
```diff
@@ -1,42 +1,256 @@
...
+        self.mediaSettings = MediaSettings(**data.find('MediaSettings').attrib)
+        self.policy = Policy(**data.find('Policy').attrib)
+        self.location = data.find('Location').attrib.get('uri', '')
 
     def server(self):
-        server = list(filter(lambda x: x.machineIdentifier == self.machineIdentifier, self._servers))
+        """ Returns :class:`~plexapi.myplex.MyPlexResource` with server of current item.
+        """
+        server = [s for s in self._server.resources() if s.clientIdentifier == self.machineIdentifier]
         if 0 == len(server):
```
Hellowlol: can you do len(server) == 0 instead,

plexapi/sync.py
```diff
@@ -1,42 +1,256 @@
...
+            any SyncItem where it presented).
+
+            Parameters:
+                media (base.Playable): the media to be marked as downloaded
+        """
+        url = '/sync/%s/item/%s/downloaded' % (self.clientIdentifier, media.ratingKey)
+        media._server.query(url, method=requests.put)
+
+    def delete(self):
+        url = SyncList.key.format(clientId=self.clientIdentifier)
```
Hellowlol: Missing doc string

plexapi/sync.py
```diff
@@ -1,42 +1,256 @@
...
+    """
+    key = 'https://plex.tv/devices/{clientId}/sync_items'
+    TAG = 'SyncList'
+
+    def _loadData(self, data):
+        self._data = data
+        self.clientId = data.attrib.get('clientIdentifier')
+        self.items = []
+
+        for elem in data:
```
Hellowlol: Can we use .iter('SyncItem') here?

plexapi/video.py
```diff
@@ -77,6 +78,31 @@ def markUnwatched(self):
         self._server.query(key)
         self.reload()
 
+    def _default_sync_title(self):
+        return self._prettyfilename()
+
+    def sync(self, client, policy, media_settings, title=None):
+        """ Add current video as sync item for specified device.
```
Hellowlol: add docs for args, and keywork arguments.

tests/test_sync.py
```diff
@@ -0,0 +1,82 @@
+from time import sleep
+
+from plexapi.sync import Policy, MediaSettings, VIDEO_QUALITY_3_MBPS_720p
+
+
+MAX_RETRIES = 3
+
+
+def get_new_synclist(account, client_id, old_sync_items=[]):
```
Hellowlol: old_sync_items is mutable, is this intended?

andrey-yantsen: I've already rewrote this part locally, anyway :D

andrey-yantsen:
Heyhey, @Hellowlol, thank you for the review! I've pushed another pile of code, where I've done few things:

1. Fixed part of your review request
2. Marked all (I hope that I marked them all) as `TODO`, to not miss it later
3. Simplified initialisation of `sync.MediaSettings` in order to get this working easily with other content types
3. Added `sync()` method to `audio`, `photo` & `playlists`
4. Each sync-related method now can accept `client` as an object, or `clientId` as a string
5. Covered this stuff with creepy tests

In order to get sync/tests for photos working PR #292 need to be applied.

Hellowlol:
Let me know when this is ready, i want to squash and merge this before i do the small prs.

andrey-yantsen:
Yep, sure. I hope to finish this by tomorrow.

Hellowlol:
Thanks for the doc changes. I havnt really looked at the code before now. There is a lot of repetitions regarding the sync methods. Anyway we could simplify this?

andrey-yantsen:
Uh... From the top of my head — I can add a mixin `Syncable` with some basic implementation of the method, but, IMO, because of different content types (video, audio, photo) require different options, and it's possible to sync so many different, not directly related in OOP to each other, stuff (library, playlist, movie, tv-show episode, etc.) there are still be some repetition... 

@Hellowlol Do you have any better ideas?

Hellowlol:
Using a mixin was my tought too be if there are too many options we might just leave it as it is. Anyway are you done?

andrey-yantsen:
Yep, it's done I think. Probably I'll find something else later, but for now I have nothing to add :)

Hellowlol:
Awesome, thanks!

Hellowlol:
Can you fix the inline imports so the style is consistent? We don't seem to use any other relative imports,

andrey-yantsen:
Yep, sure, I've just pushed it

PR #288, opened 2018-09-01
Add enable_fast_connect configuration parameter

andrey-yantsen:
Most times I run the my things outside my home network, when the local IP of the server is not available, and it takes annoyingly long time to establish the connection. With this option set you don't have to wait until all available connection methods will be established.

Moved from #282.

### DIFF
--- a/plexapi/__init__.py
+++ b/plexapi/__init__.py
@@ -17,6 +17,7 @@ PROJECT = 'PlexAPI'
 VERSION = '3.0.6'
 TIMEOUT = CONFIG.get('plexapi.timeout', 30, int)
 X_PLEX_CONTAINER_SIZE = CONFIG.get('plexapi.container_size', 100, int)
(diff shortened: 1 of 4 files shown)


plexapi/__init__.py
```diff
@@ -17,6 +17,7 @@
 VERSION = '3.0.6'
 TIMEOUT = CONFIG.get('plexapi.timeout', 30, int)
 X_PLEX_CONTAINER_SIZE = CONFIG.get('plexapi.container_size', 100, int)
+X_PLEX_ENABLE_FAST_CONNECT = CONFIG.get('plexapi.enable_fast_connect', False, bool)
```
Hellowlol: Any downside to let this just be enabled by default?

andrey-yantsen: There is a chance that you'll connect to an external IP while you're in your home network. It may be a bit unexpected and strange.

PR #290, opened 2018-09-02
Bootstrap test plex

andrey-yantsen:
I've created a script `tools/plex-bootstraptest.py` which creates a fresh Plex instance in a docker container (by calling the `docker` executable). Here is the details on it's workflow:

1. Check if docker available, quit otherwise
2. Check if any IP is available except local-loop, quit otherwise
3. Pull fresh instance of `plexinc/pms-docker`
4. Get "claim token". Basically MyPlexAccount is required only because of this step, but anyway we need the account to test everything properly
5. Create a plex instance, wait until it become available
6. Prepare folders structure for new library
7. Download Philips Circle Test Card, if not exists already, from http://www.mytvtestpatterns.com/, to use it for all required videos
8. Create 4 movies and 2 TV Shows as stated in tests/conftest.py, all videos are created as hard links (or complete copies when using macOS) to the Test Card
9. Download `Infinite State - Unmastered Impulses` and `Broke For Free - Layers` for Music library
10. Download 10 images from https://picsum.photos/
11. Query an update on each library
12. When all required photos become available — create a photo album `photo_album1` (no cats, sorry)
13. As soon as all media scanned by Plex the script terminates.

P.S. I have to do a quite dirty hack in 3787596, I have no idea how else I could receive a librarySectionID for photos, which are not in album yet.
P.P.S. Copied this hack into separate PR #292.

### DIFF
--- a/plexapi/compat.py
+++ b/plexapi/compat.py
@@ -2,6 +2,7 @@
 # Python 2/3 compatability
 # Always try Py3 first
 import os
+import sys
 from sys import version_info
 
 ustr = str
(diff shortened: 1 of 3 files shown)


andrey-yantsen:
Also I need your advice about photos: I can't find "good old albums" in current Plex, aren't they were completely replaced with playlists?

PR #292, opened 2018-09-03
Library section id for all

andrey-yantsen:
It's a quite controversial change, but for now I have no idea how to do this thing better. Basically `librarySectionID` field should be added automatically for any item if the field exists in metadata, and that's what I'm trying to achieve here.

### DIFF
--- a/plexapi/base.py
+++ b/plexapi/base.py
@@ -145,7 +145,12 @@ class PlexObject(object):
             on how this is used.
         """
         data = self._server.query(ekey)
-        return self.findItems(data, cls, ekey, **kwargs)
+        items = self.findItems(data, cls, ekey, **kwargs)
+        librarySectionID = data.attrib.get('librarySectionID')
+        if librarySectionID:
+            for item in items:
+                item.librarySectionID = librarySectionID
+        return items
 
     def findItems(self, data, cls=None, initpath=None, **kwargs):
         """ Load the specified data to find and build all items with the specified tag
@@ -159,13 +164,10 @@ class PlexObject(object):
             kwargs['type'] = cls.TYPE
         # loop through all data elements to find matches
         items = []
-        librarySectionID = data.attrib.get('librarySectionID')
         for elem in data:
             if self._checkAttrs(elem, **kwargs):
                 item = self._buildItemOrNone(elem, cls, initpath)
                 if item is not None:
-                    if librarySectionID:
-                        item.librarySectionID = librarySectionID
                     items.append(item)
         return items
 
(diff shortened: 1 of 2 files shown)


Hellowlol:
Why Is this needed?

andrey-yantsen:
@Hellowlol there are photos in a library which not belong to any albums (and probably albums are not a thing anymore), and for those photos you can't get a `librarySectionID` in a nice&easy way. Without the field you can't create playlist with those photos, or send them to sync.

At least `Photo.section()` fails when the photo received from `PhotoSection.all()`.

andrey-yantsen:
Problem occurs at least on with PMS v1.13.5.5332. Of course there is a huge chance that I'm doing anything wrong, but I really hope that it's not the case this time :)

andrey-yantsen:
Here is a short piece of code (from py.test) which fails with exception about unknown library:
```python
def test_playlist_photos(plex, photos):
    items = photos.all()
    assert plex.createPlaylist('Photos', items)
```

Hellowlol:
Im gonna put this on hold for a while, i want to see if we can do this another way. This touches parts of code that needs to be fast as its used pretty much everywhere.

andrey-yantsen:
> Im gonna put this on hold for a while, i want to see if we can do this another way.

Yep, I really hope we find a better way :) But anyway — when we receiving a MediaContainer IMO we should parse all of its attributes, not just processing internal items and abandoning all the meta. E.g. for `library.section('Photos').all()` I'm receiving following meta:

```xml
<MediaContainer allowSync="1" art="/:/resources/photo-fanart.jpg" identifier="com.plexapp.plugins.library" librarySectionID="4" librarySectionTitle="Photos" librarySectionUUID="2a550866-7060-4612-9f64-2b7b4e0a82b7" mediaTagPrefix="/system/bundle/media/flags/" mediaTagVersion="1534566129" nocache="1" size="10" thumb="/:/resources/photo.png" title1="Photos" title2="All Photos" viewGroup="photo" viewMode="131633">
```

And at least 3 attributes are pretty much useful for me: `allowSync`, `librarySectionID` and `librarySectionUUID`. Also `librarySectionUUID` would be helpful when creating new playlists. Looks like `fetchItems` should return a some kind of `MediaContainer` object, exactly as it is in the API.

> Who creates a playlist for images anyway? :P

In latest version of Plex "playlist for images" used as albums :( And also you can't sent selection of photos without library ID to sync.

@Hellowlol I've made another try on this, this time it doesn't have this huge side-effect, please take a look on 5d9df1a.

andrey-yantsen:
Yep. Without this one you can't properly work (create playlists, set up mobile sync) with photos which stored outside of any albums.

Hellowlol:
Still dont like it but lets go for it.

PR #294, opened 2018-09-06
Passed kwargs into constructor to allow verify=False to be used by re…

barrycarey:
If the PMS doesn't have a valid SSL requests fails to connect to the server. 

HTTPSConnectionPool is thrown if using self signed cert. 

This request allows verify=False to be passed to the PlexServer constructor

### DIFF
--- a/plexapi/server.py
+++ b/plexapi/server.py
@@ -91,7 +91,7 @@ class PlexServer(PlexObject):
     """
     key = '/'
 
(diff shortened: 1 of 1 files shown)


pkkid:
This should be done by modifying the session when initializing PlexServer.

Hellowlol:
See https://github.com/pkkid/python-plexapi/issues/295#issuecomment-418991008 example use.

PR #297, opened 2018-09-08
Improvements in tests process

andrey-yantsen:
Currently I can see following problems with tests:

1. [X] Tests which run on Travis-CI depends on external Plex Server
2. [X] The tests aren't running for pull requests
3. [X] flake8 executed for each python version, which is senselessly
4. [X] Deploy to PyPi executed for each python version which lead to errors in uploading
5. [X] The tests require quite a work on building a library, but even after the library is ready not all tests are passing
6. [x] Part of tests are require an existence of user `PKKid`
7. [x] Some of `MyPlexAccount` test require that you don't have a PlexPass (and it's not really documented)
8. [x] My Mobile Sync tests require a Plex Pass on other hand, and this is also not documented
9. [x] The readme not covers a testing procedure
10. [x] Test results should be reproducible, without "unstable" tests
11. [x] Tests agains shared libraries should be easy-to-run, and the should run automatically in Travis 

As for now I've partially resolved first few easy issues, and stuck with generating a proper library: I have no idea how to get tests related to `photoalbum` to pass, and I need your help here.

To get my new approach working you have to define `PLEXAPI_AUTH_MYPLEX_USERNAME` and `PLEXAPI_AUTH_MYPLEX_PASSWORD` in the your Travis CI ENV variables, instead of `PLEXAPI_AUTH_SERVER_BASEURL` and `PLEXAPI_AUTH_SERVER_TOKEN`.

After each run of travis the newly created server and client would be removed from the MyPlex account.

You can check the Travis CI output by visiting https://travis-ci.org/andrey-yantsen/python-plexapi/builds/426206148.

### DIFF
--- a/plexapi/server.py
+++ b/plexapi/server.py
@@ -293,8 +293,6 @@ class PlexServer(PlexObject):
         releases = self.fetchItems('/updater/status')
         if len(releases):
             return releases[0]
-        else:
-            return None
 
     def isLatest(self):
         """ Check if the installed version of PMS is the latest. """
--- a/plexapi/server.py
+++ b/plexapi/server.py
@@ -299,7 +299,6 @@ class PlexServer(PlexObject):
     def isLatest(self):
         """ Check if the installed version of PMS is the latest. """
         release = self.check_for_update(force=True)
-        print(release)
         return release is None
 
     def installUpdate(self):
--- a/plexapi/server.py
+++ b/plexapi/server.py
@@ -290,12 +290,17 @@ class PlexServer(PlexObject):
         part = '/updater/check?download=%s' % (1 if download else 0)
         if force:
             self.query(part, method=self._session.put)
-        return self.fetchItem('/updater/status')
+        releases = self.fetchItems('/updater/status')
+        if len(releases):
+            return releases[0]
+        else:
+            return None
 
     def isLatest(self):
         """ Check if the installed version of PMS is the latest. """
         release = self.check_for_update(force=True)
-        return bool(release.version == self.version)
+        print(release)
+        return release is None
 
     def installUpdate(self):
         """ Install the newest version of Plex Media Server. """
--- a/plexapi/video.py
+++ b/plexapi/video.py
@@ -323,7 +323,7 @@ class Show(Video):
 
     def seasons(self, **kwargs):
         """ Returns a list of :class:`~plexapi.video.Season` objects. """
-        key = '/library/metadata/%s/children' % self.ratingKey
+        key = '/library/metadata/%s/children?excludeAllLeaves=1' % self.ratingKey
         return self.fetchItems(key, **kwargs)
 
     def season(self, title=None):
--- a/tests/test_myplex.py
+++ b/tests/test_myplex.py
@@ -109,7 +109,6 @@ def test_myplex_deletewebhooks(account):
             account.deleteWebhook('http://example.com')
 
 
-@pytest.mark.flaky(reruns=5, reruns_delay=2)
 def test_myplex_optout(account_once):
     def enabled():
         ele = account_once.query('https://plex.tv/api/v2/user/privacy')
--- a/tests/test_myplex.py
+++ b/tests/test_myplex.py
@@ -152,3 +152,12 @@ def test_myplex_updateFriend(account, plex, mocker):
                                  allowCameraUpload=True, allowChannels=False, filterMovies=vid_filter,
                                  filterTelevision=vid_filter, filterMusic={'label': ['foo']})
 
+
+def test_myplex_plexpass_attributes(account_plexpass):
+    assert account_plexpass.subscriptionActive
+    assert account_plexpass.subscriptionStatus == 'Active'
+    assert account_plexpass.subscriptionPlan
+    assert 'sync' in account_plexpass.subscriptionFeatures
+    assert 'premium_music_metadata' in account_plexpass.subscriptionFeatures
+    assert 'plexpass' in account_plexpass.roles
+    assert 'all' in account_plexpass.entitlements
--- a/tests/test_myplex.py
+++ b/tests/test_myplex.py
@@ -98,9 +98,9 @@ def test_myplex_deletewebhooks(account):
         account.deleteWebhook('http://site.com')
 
 
-def test_myplex_optout(account):
+def test_myplex_optout(account_once):
     def enabled():
-        ele = account.query('https://plex.tv/api/v2/user/privacy')
+        ele = account_once.query('https://plex.tv/api/v2/user/privacy')
         lib = ele.attrib.get('optOutLibraryStats')
         play = ele.attrib.get('optOutPlayback')
         return bool(int(lib)), bool(int(play))
@@ -108,11 +108,11 @@ def test_myplex_optout(account):
     # This should be False False
     library_enabled, playback_enabled = enabled()
 
-    account.optOut(library=True, playback=True)
+    account_once.optOut(library=True, playback=True)
 
     assert all(enabled())
 
-    account.optOut(library=False, playback=False)
+    account_once.optOut(library=False, playback=False)
 
     assert not all(enabled())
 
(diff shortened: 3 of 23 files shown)


andrey-yantsen:
Which version of Plex are you currently using for tests? There is no setting named `collectUsageData` in latest public build.

Hellowlol:
Dunno, Its @pkkid test server.

Hellowlol:
I dont know. @pkkid what version is the test server? Our test account don’t have plex pass so these need to be be skipped if plex pass is missing.

Hellowlol:
Dunno, my suggestion would be to build a root dir with tvshow, movies, photo and music. Add the files there and just zip them. We could download and setup the entire thing on demand. I’m a bit worried about testing time and I don’t want to keep downloading the same files over and over. Metadata creation might take some time too.

andrey-yantsen:
Testing time is ~8 minutes (or 20 minutes, if you test all python versions sequentially) with bootstrapping from scratch: download the required media, set up libraries, wait for metadata, and this is exactly how the tests working in this PR. It's about 4 times longer the the your current approach.

andrey-yantsen:
Ok, looks like I've resolved the issue with photos: Plex have 2 types of photo albums:
1. Playlists
2. Physical folders on the server

And I initially found only first one. Thank you for the attention :) I hope other issues would be less problematic.

pkkid:
@andrey-yantsen - I'm OK with tests taking 4 times longer if we are standing up and tearing down the Plex server as part of it. It's a much more reliable way to test moving forward. -- We should add the requirements to get tests working to the README of the project. I think all it requires beyond the pip install is the environment variables mentioned above?

andrey-yantsen:
Ok guys, I've achieved some progress on this today, you can check the build results on my [travis](https://travis-ci.org/andrey-yantsen/python-plexapi/builds/426412786) account.

@pkkid @Hellowlol What actually tests `test_myplex_users`? It's the last failing test (except skipped ones, due do some dependencies).

Changes in tests (and travis) so far:

0. You have to provide `PLEXAPI_AUTH_MYPLEX_USERNAME` and `PLEXAPI_AUTH_MYPLEX_PASSWORD` as environment variables (or set MyPlex auth data by any other convenient way)
1. Plex instance created for each tests run on Travis, each time with proper library
2. After the test (either successful or failed) the server (as well as plexapi-client) is removed from MyPlex account
3. You can create a local copy of required Plex instance using `tools/plex-bootstraptest.py` and docker, it'll download all required video/music/photos and create all required things
4. For tests which require a shared user you can specify the username in env variable `SHARED_USERNAME`, if the user wasn't found — test will be skipped, default value is `PKKid`
5. Some tests were fixed for actual versions of Plex Server
6. Travis now run tests 4 times:
    1. Python 2.7, Plex Latest
    2. Python 3.4, Plex Latest
    3. Python 3.6, Plex Latest
    4. Python 3.6, Plex 1.3.2, also in this iteration `MyPlexAccount.optOut` is tested, to prevent interference with other runs. Also this step required to properly test `PlexServer.isLatest()`
7. `websocket-client` locked to version `0.48.0` due to websocket-client/websocket-client#471 (seems like the issue is fixed now, I'll test it later)
8. Flake8 runs only once and not for each python version as previously
9. Deploy script will be ran only once, if tag was provided, to prevent build failures on release.

And the following things left to do:
1. Fix `test_myplex_users` (or skip it with proper conditions)
2. Add testing description to readme
3. Fix `test_myplex_optout`: it randomly fails, looks like some kind of cache on plex.tv side
4. Check if tests against shared library are working

I hope I'll be able to finish all of this on next week (but if you feel like joining me — feel free to do so, seems like next couple weeks would be quite busy for me), and all of us could finally receive a new release :)

andrey-yantsen:
Ok, never mind, I've fixed `test_myplex_users` as I understood it, I hope it's the right way :) Here is an all-green build: https://travis-ci.org/andrey-yantsen/python-plexapi/builds/426418368. Unfortunately it's not "clear" green — I had to restart the build a couple times, because some tests aren't really stable.
Anyway, 🎉🎉🎉! :)

Hellowlol:
We could use 
https://github.com/pytest-dev/pytest-rerunfailure on flaky tests

andrey-yantsen:
Yep, for things like `test_myplex_optout` sounds like a good & easy idea, thank you. I've just added it, and now can't get other flaky tests to fail to detect them :D

andrey-yantsen:
Do you thing it is really ok to rerun all failed tests? I thought about just marking the flanky ones and rerun only them.

pkkid:
I don't think it's right to rerun any tests. If a test is flaky, rerunning it is just rolling the dice again. My vote would be rerun nothing and fix the test or toss it out.

andrey-yantsen:
Ok guys, seems like I've done here, please review all of this.

As for websocket-client — version 0.53.0 didn't worked on Travis from the first try and I didn't investigated it further.

tests/test_myplex.py
```diff
@@ -87,23 +87,23 @@ def test_myplex_resource(account, plex):
 
 def test_myplex_webhooks(account):
     if account.subscriptionActive:
-        assert not account.webhooks()
+        assert type(account.webhooks()) is list
```
Hellowlol: why not isinstance?

plexapi/server.py
```diff
@@ -290,12 +290,16 @@ def check_for_update(self, force=True, download=False):
         part = '/updater/check?download=%s' % (1 if download else 0)
         if force:
             self.query(part, method=self._session.put)
-        return self.fetchItem('/updater/status')
+        releases = self.fetchItems('/updater/status')
+        if len(releases):
+            return releases[0]
+        else:
```
Hellowlol: the else part could just be removed. since returns None anyway.

plexapi/video.py
```diff
@@ -323,7 +323,7 @@ def isWatched(self):
 
     def seasons(self, **kwargs):
         """ Returns a list of :class:`~plexapi.video.Season` objects. """
-        key = '/library/metadata/%s/children' % self.ratingKey
+        key = '/library/metadata/%s/children?excludeAllLeaves=1' % self.ratingKey
```
Hellowlol: Whats the benefit by excludeAllLeaves? Faster?

andrey-yantsen: Without it we'd have "All Seasons" as one more seasons, I think it's a bit unexpected.

Hellowlol: Im not sure i understand. This is supposed to returns all seasons. does that still happen?

andrey-yantsen: Yep, sure. If the show has 2 seasons it'll return [<Season 1>, <Season 2>], but if you'd omit `excludeAllLeaves` you'll receive [<All Seasons>, <Season 1>, <Season 2>].

andrey-yantsen:
Oh, guys, just remembered: I know nothing about client-related tests, can you please cover it in the docs?

Hellowlol:
i dont think the clients is tested at all atm as it needs to be on the same local network as the server. Really stellar work btw! Thanks you so much.

andrey-yantsen:
Ooooh, I have one more idea, I can't stop it, sorry :) I'll try to automate the client tests this evening using Plex Web and some Selenium magic.

andrey-yantsen:
I've made some progress with client tests, but in order to get them working I need to dig into PlexClient and rewrite some parts of it. You can take a look at my working branch for this, if you're interested: https://github.com/andrey-yantsen/python-plexapi/tree/client-tests.

Shortly: In my working branch I was able to create an instance of Plex Web within a docker, and also I was able to play a photo/video on it using my real browser, but that's all what I've achieved.

I'll perform another attempt later, but I think it would be better to do it within another PR.

andrey-yantsen:
@Hellowlol you need to remove `PLEXAPI_AUTH_SERVER_TOKEN` from env in Travis

andrey-yantsen:
image
🎉 🎉 🎉

PR #298, opened 2018-09-10
[WIP] [Do not merge] Add support for mobile providers (news & podcasts)

andrey-yantsen:
Plex recently added podcasts support, I think it would be useful to be able to manage them and be able to download required episodes. And because Podcasts & News uses similar API, I'd like to add them both.

Currently I've faced 2 problems:

1. Request to `https://plex.tv/media/providers` seems to require `X-Plex-Product` to be set to one of official clients (with proper `X-Plex-Version`), because it returns empty list otherwise
2. `ratingKey` for podcasts returned as hexadecimal value (e.g. `5a976ae1a73f0c0012c15392`), not plain int. It's not really a problem, but cat break some compatibility if implemented not carefully enough (like what I've done for now :D)
3. Parts for media in News provider returned as `video/mpd` and not as real videos (seems like we could use a library https://github.com/onovy/python-mpd-parser as a quickstart, I haven't investigated yet).

As for `ratingKey` — there is a `subtype` tag in each `track` (for podcasts) and `clip` (for news), probably we could use it to select proper classes, which would use string-ish ratingKey.


andrey-yantsen:
Doesn't seems like I'll continue work on this PR in near time, sorry.

PR #302, opened 2018-09-15
Run tests over unclaimed server

andrey-yantsen:
I've added some extra checks and now we have an additional run of tests in Travis: over unclaimed server. You can check out how it looks on my fork's build: https://travis-ci.org/andrey-yantsen/python-plexapi/builds/429074802.

With this PR tests will be ran over PRs, they'll automatically run with unclaimed server.

Also, since we're not using an external server for testing anymore, maybe we can disable [Limit concurrent jobs](https://docs.travis-ci.com/user/customizing-the-build/#limiting-concurrent-jobs) in travis? I'm guessing it's enabled now, because the jobs are running one-by-one, and with this scheme the tests (5 runs of them) would be taking about an hour to complete.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,7 @@
 language: python
 
 stages:
-  - syntax
-  - name: test
-    if: type != pull_request
+  - test
   - name: deploy
     if: tag = present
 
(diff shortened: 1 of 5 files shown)


andrey-yantsen:
I don't know what I can do about the coverage report: with claimed server we have 184 tests, but with unclaimed one we have only 137 tests, of course the coverage will be lower...

andrey-yantsen:
@Hellowlol @pkkid guys, what do you think about this one?

Hellowlol:
I guess some tests for prs are better then none. Will the normal tests be run on on normal commits?

pkkid:
I love the idea of testing being more reliable. I just haven't had time to look at or play with the code yet to merge it in. If Hellowlol thinks it looks good, I'm fine merging it.

Can you explain [these results](https://travis-ci.org/pkkid/python-plexapi/jobs/429176682) a little more? I am seeing the tests showing [authenticated] and [anonymous]. Is it correct that we can only test anonymous things on a PR, but when committing to master it will test [authenticated] only or both?

andrey-yantsen:
`authenticated` runs only on non-PR commits (because Travis hides authentication data from PRs), `anonymous` runs for all commits — in PRs and in master, but unfortunately in current implementation they have to run separately, that's why I created another testing step in travis config (which runs only for non-PR).

You can check out how it would look on commits to master in a travis-build of my fork: [authenticated](https://travis-ci.org/andrey-yantsen/python-plexapi/jobs/429074805) and [anonymous](https://travis-ci.org/andrey-yantsen/python-plexapi/jobs/429074808).

pkkid:
Do we need to update any Travis config variables after merging, or just merge and we're good?

andrey-yantsen:
It will be working right after the merge, but IMO it's good idea to disable [Limiting Concurrent Jobs](https://docs.travis-ci.com/user/customizing-the-build/#limiting-concurrent-jobs), just to speed up the build.

andrey-yantsen:
> More questions.. Do I need to maintain my testing plex server anymore, or can that server be trashed?

I'm not sure, actually... Probably you can trash it, but I still hadn't took a look properly onto client tests, so I don't know their requirements.

pkkid:
Most client tests required manual runs against a phone, PHT, or PlexWeb interface to really work, so I'm fine if these stay that way for now. Thanks again!

andrey-yantsen:
All tests has been passed in the [last build](https://travis-ci.org/pkkid/python-plexapi/builds/437136914) in 7 minutes 🎉

PR #303, opened 2018-09-20
additional attributes

blacktwin:
target, title and video profile.  Optimizing items through Plex now allows for adding a title to the version. Target and title are the version name and seem to be the same value.


Hellowlol:
Mind adding docs string for video profile?

PR #329, opened 2019-01-07
Fix spelling of "original"

hmnd:
Fix spelling of "original", formerly spelled "orignal".


Hellowlol:
Thanks for the PR, im gonna hold off on this as this might break things for users. Ill add it to a breaking release :)

PR #332, opened 2019-01-19
fix datetime issue on windows.

Hellowlol:


pkkid:
Uhhg, what a silly bug.

PR #335, opened 2019-02-04
Add 'forced' var to subtitleStream

gstacks13:
Add a 'forced' variable to subtitleStream to access the forced flag from the xml.


pkkid:
Cool, thanks!

PR #336, opened 2019-02-04
Closes #334 - set default audio & subtitle stream

### LINKED ISSUE #334: Add a way to set default audio and subtitle stream.
Preset audio stream can be set by sending `PUT` request to `/library/parts/<PART_ID>?audioStreamID=<AUDIO_STREAM_ID>&allParts=1` and subtitles by `PUT` to `/library/parts/<PART_ID>?subtitleStreamID=<SUBTITLES_STREAM_ID>&allParts=1`. You can get all the placeholders from your `Media`-object, to be precise — `.parts` field of it, and then send the query with `server.query(URL, method=server._session.put)`.

_Originally posted by @andrey-yantsen in https://github.com/pkkid/python-plexapi/issues/333#issuecomment-459638015_

gstacks13:
Set the 'selected' flags of audio & subtitle streams of a MediaPart. #334.

### DIFF
--- a/plexapi/media.py
+++ b/plexapi/media.py
@@ -125,22 +125,42 @@ class MediaPart(PlexObject):
         """ Returns a list of :class:`~plexapi.media.SubtitleStream` objects in this MediaPart. """
         return [stream for stream in self.streams if stream.streamType == SubtitleStream.STREAMTYPE]
         
-    def setDefaultAudioStream(self, id):
+    def setDefaultAudioStream(self, stream=None, streamID=None):
         """ Set the default :class:`~plexapi.media.AudioStream` for this MediaPart.
 
             Parameters:
-                id (int): ID of the AudioStream to set
+                stream (:class:`~plexapi.media.AudioStream`): AudioStream to set as default 
+                    (default:None; required if streamID is not specified).
+                streamID (int): ID of the AudioStream to set 
+                    (default:None; required if stream is not specified).
+            
+            Raises:
+                :class:`plexapi.exceptions.BadRequest`: If both stream and streamID are missing.
         """
-        key = "/library/parts/%d?audioStreamID=%d&allParts=1" % (self.id, id)
+        if stream:
+            key = "/library/parts/%d?audioStreamID=%d&allParts=1" % (self.id, stream.id)
+        elif streamID:
+            key = "/library/parts/%d?audioStreamID=%d&allParts=1" % (self.id, streamID)
+        else:
+            raise BadRequest('Missing argument: stream or streamID is required')
         self._server.query(key, method=self._server._session.put)
-        
-    def setDefaultSubtitleStream(self, id):
+            
+    def setDefaultSubtitleStream(self, stream=None, streamID=None):
         """ Set the default :class:`~plexapi.media.SubtitleStream` for this MediaPart.
+            (Note: pass no parameters to disable subtitles)
 
             Parameters:
-                id (int): ID of the SubtitleStream to set (0 for no subtitles)
+                stream (:class:`~plexapi.media.SubtitleStream`): SubtitleStream to set as default 
+                    (default:None).
+                streamID (int): ID of the AudioStream to set 
+                    (default:None).
         """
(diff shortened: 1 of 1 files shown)


Hellowlol:
What ID is passed here? Can we also pass the object

gstacks13:
The ID for the audioStream & subtitleStream is passed, e.g., `part.setDefaultAudioStream(audioStream.id)`. I can change it to pass the object if that would make it more clear.

Hellowlol:
Just make it accept both :)

gstacks13:
Sure thing :)

Hellowlol:
also id is a builtin can you change it to something else

gstacks13:
Can now pass both audioStream and audioStream.id successfully. Same for subtitles. Let me know if anything else needs adjusted!

andrey-yantsen:
Maybe it's better to create separate method for reseting subtitles? IMO it's not quite obvious that you need to pass 0 as an id to reset it, even if it's defined in the docs...

Hellowlol:
Just pass the object. If it’s the correct class use the class.id or the id.

gstacks13:
@Hellowlol I agree - it's much more clear to just pass the object and use the ID from there.  I'll get that change up later this evening.

@andrey-yantsen I have it now where it'll disable subtitles if you pass no parameters. I.e., `part.setDefaultSubtitleStream()` will disable subs.  Do you still think that isn't clear enough?

plexapi/media.py
```diff
@@ -124,7 +124,41 @@ def audioStreams(self):
...
+    def setDefaultAudioStream(self, stream):
+        """ Set the default :class:`~plexapi.media.AudioStream` for this MediaPart.
+
+            Parameters:
+                stream (:class:`~plexapi.media.AudioStream`): AudioStream to set as default 
+            
+            Raises:
+                :class:`plexapi.exceptions.BadRequest`: If stream is not an AudioStream.
+        """
+        if type(stream) == AudioStream:
```
andrey-yantsen: I think it's better to use `isinstance()` here, instead of direct types comparison, just in case if somebody suddenly create own implementation of `AudioStream`

plexapi/media.py
```diff
@@ -124,7 +124,41 @@ def audioStreams(self):
...
+            key = "/library/parts/%d?audioStreamID=%d&allParts=1" % (self.id, stream.id)
+            self._server.query(key, method=self._server._session.put)
+        else:
```
andrey-yantsen: I believe that @Hellowlol have meant that you need to use `stream` value as in ID if it's not an `AudioStream`-object

plexapi/media.py
```diff
@@ -124,7 +124,41 @@ def audioStreams(self):
...
+    def setDefaultSubtitleStream(self, stream):
+        """ Set the default :class:`~plexapi.media.SubtitleStream` for this MediaPart.
+            
+            Parameters:
+                stream (:class:`~plexapi.media.SubtitleStream`): SubtitleStream to set as default.
+            
+            Raises:
+                :class:`plexapi.exceptions.BadRequest`: If stream is not a SubtitleStream.
+        """
+        if type(stream) == SubtitleStream:
```
andrey-yantsen: Same here about types comparison and using `stream` as an id

plexapi/media.py
```diff
@@ -124,7 +124,41 @@ def audioStreams(self):
...
+            key = "/library/parts/%d?subtitleStreamID=%d&allParts=1" % (self.id, stream.id)
+            self._server.query(key, method=self._server._session.put)
+        else:
+            raise BadRequest("Object 'stream' is not a SubtitleStream.")
+        
+    def resetSubtitles(self):
```
andrey-yantsen: Can you please rename the method a little? Currently we have `setDefaultSubtitleStream` and `resetSubtitles`, which looks a bit strange — I believe that the second one should be `resetDefaultSubtitleStream`

gstacks13:
Not being a pain in the ass - this is all very constructive!  I made the changes you requested - let me know if there's anything else that needs adjusting!

pkkid:
Thanks for the updates.

PR #339, opened 2019-02-07
Fix Show.season(0) not returning 'Specials'

gstacks13:
Currently Show.season(0) attempts to return a season with the title 'Season 0', which throws an exception because Plex titles Season 0 as 'Specials'.  This fixes that bug.

### DIFF
--- a/plexapi/video.py
+++ b/plexapi/video.py
@@ -333,7 +333,7 @@ class Show(Video):
                 title (str or int): Title or Number of the season to return.
         """
         if isinstance(title, int):
-            title = 'Season %s' % title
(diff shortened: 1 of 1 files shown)


gstacks13:
Is there any reason this is failing the build tests?  It looks like it's failing creating smart playlists, which I didn't even modify.

Hellowlol:
It’s a old error this is fixed in master. But I’m on mobile and isn’t allowed to merge without all green

gstacks13:
Oh no worries at all. Just making sure it wasn't anything on my end.

PR #341, opened 2019-02-10
Fix show.episode() not allowing season 0

gstacks13:
In show.episode(), a season value of 0 will evaluate to false, and the function thinks a season has not been provided.  Since season 0 is a valid Plex season, this now explicitly checks for a value of 'None', which correctly evaluates season 0 as valid.


Hellowlol:
Thanks!

PR #343, opened 2019-02-21
Get season from index instead of creating title

gstacks13:
Currently, when passing a season number X, this function creates a season title "Season X", then fetches the season based off that. If the user has changed the title of their season to something more befitting of the show, this obviously fails. This patch fixes that by grabbing a season based off the Season.index when an integer is passed.

### DIFF
--- a/plexapi/video.py
+++ b/plexapi/video.py
@@ -332,9 +332,9 @@ class Show(Video):
             Parameters:
                 title (str or int): Title or Number of the season to return.
         """
-        if isinstance(title, int):
-            title = 'Season %s' % title if title != 0 else 'Specials'
         key = '/library/metadata/%s/children' % self.ratingKey
+        if isinstance(title, int):
+            return self.fetchItem(key, etag='Directory', index__iexact=str(title))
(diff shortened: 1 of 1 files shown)


andrey-yantsen:
Nice catch, thanks! Can you update the tests as well? Currently the test library bootstraps using [plex-bootstraptest.py](https://github.com/pkkid/python-plexapi/blob/master/tools/plex-bootstraptest.py) script, I assume we need to add one more show, containing `Specials` season and another one, with changed name. Don't hesitate to poke me if you have any questions about testing env.

Hellowlol:
This is already tested in 
https://github.com/pkkid/python-plexapi/blob/master/tests/test_video.py#L564

andrey-yantsen:
Yep, but we don't have season #0 here, and custom-named seasons.

Hellowlol:
True. But the pr stills test the same as before. Ideally we should have a season 0.

gstacks13:
Thanks guys! Let me know about the tests - I'd be happy to update them if need be.

PR #348, opened 2019-05-22
Don't warn when also raising exception

balloob:
We should not log warnings if we are also raising exceptions. It is up to the caller to decide how to handle the error and if they should notify the user. 

This was triggered because I got logs for bad requests for 401 Unauthorized.


pkkid:
Agreed, thanks!

PR #351, opened 2019-05-29
added collection to show object for tagging

pbrink231:
I need to add collection tags to a show.

This edit has worked in my test.


pbrink231:
The checks looks like there is an issue with the pictures library in your testing bed.

The code should have no effect on those.  Also this worked no problem for my app.

pkkid:
thanks

Issue #350, opened 2019-05-24
Family and Friends settings exposed


MichaelFay:
It would be very useful if there was the ability to create/delete a home user.

MichaelFay:
The parameters will need to change to something like this: 

```python
params = {
            'server_id': machineId,
            'shared_server': {'library_section_ids': sectionIds, **'title'**: username},
            'sharing_settings': {
                'allowSync': ('1' if allowSync else '0'),
                'allowCameraUpload': ('1' if allowCameraUpload else '0'),
                'allowChannels': ('1' if allowChannels else '0'),
                'filterMovies': self._filterDictToStr(filterMovies or {}),
                'filterTelevision': self._filterDictToStr(filterTelevision or {}),
                'filterMusic': self._filterDictToStr(filterMusic or {}),
            },
        }

```

blacktwin:
Have you tried out this code? Do you intend to create a PR? 

Looks like its a copy and paste of the `inviteFriend()` function with the exception to the `invited_email` to `title`. 

I think the `FRIENDINVITE` url would be incorrect. The url should be `plex.tv/api/home/users?title={title}` not `'plex.tv/api/servers/{machineId}/shared_servers'` although when creating a home user that url is posted as well but with `invited_id` instead of `invited_id`. So it looks like there would be need for additional code to be added.

MichaelFay:
@blacktwin I can definitely pull the project and test that functionality then create a pr. I won't be able to get to it for a little bit as some stuff has come up.

blacktwin:
@Hellowlol @pkkid I believe this can now be closed.

PR #365, opened 2019-07-25
Adding Managed User support

blacktwin:
In response to #350 

Adding ability to create and remove a Home User. Adding ability to create an Existing Plex User. When creating an Existing Plex User that already exists in Sharing/Friend, carry over section shares and settings. If Existing Plex User does not already exist in Sharing/Friend, sections and settings can be applied.

### DIFF
--- a/plexapi/myplex.py
+++ b/plexapi/myplex.py
@@ -62,6 +62,7 @@ class MyPlexAccount(PlexObject):
             _session (obj): Requests session object used to access this client.
     """
     FRIENDINVITE = 'https://plex.tv/api/servers/{machineId}/shared_servers'                     # post with data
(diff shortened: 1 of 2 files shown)


Hellowlol:
I'm not really sure what the createExistingUser, what the differance between this on the old share method?

blacktwin:
@Hellowlol createExistingUser will move an existing remote/Friend into your Home as a Managed User. Stole the name from the WebUI.

I'll add tests. Is there an assigned dumbie account that the test environment can add/remove/edit?

Hellowlol:
There is fixure called shared_username. You should checkout how the invite friend test works

blacktwin:
Tests should be good. A Home User will be created and removed. An Existing User will be invited in the home and then the invite is removed.

PR #366, opened 2019-07-25
Add collectionMode data and modeUpdate

blacktwin:
Added collectionMode to collection data. 
modUpdate acts like webUI, `collection.modeUpdate(mode="default")`


Hellowlol:
Can you add a test and docs string?

blacktwin:
docstrings are added. Not sure about what you'd want for the tests. I'll look into it.

blacktwin:
I'm guessing this fails because there are either no Collections or they aren't enabled in the test environment. Should I add a test to create a collection? test_actions or test_library?

Hellowlol:
I would prefer it was added to the class you have added it to. Dunno why exists a action. I have tried to follow a format when i added the initial tests. test_pyfile_Classname_method

blacktwin:
I think this is one is ready.

PR #367, opened 2019-07-26
Do not fail while parsing illegal dates. Return None instead.

makeroo:
hi,

I'm using your project and I stumbled upon a nasty Plex db having malformed dates here and there.
This caused ValueError exceptions, while parsing Plex XML, that emerged without any context.
Example: audio.Artist.albums() method fails if one album has illformed originallyAvailableAt attribute.

I'm not fond of the solution I propose you, but I needed a quick patch and the new behaviour is a bit better, at least for me.

### DIFF
--- a/plexapi/utils.py
+++ b/plexapi/utils.py
@@ -9,11 +9,8 @@ from datetime import datetime
 from getpass import getpass
 from threading import Thread, Event
 from tqdm import tqdm
-from plexapi import compat
+from plexapi import compat, log
 from plexapi.exceptions import NotFound
-
-
-LOG = logging.getLogger(__name__)
 
 
 # Search Types - Plex uses these to filter specific media types when searching.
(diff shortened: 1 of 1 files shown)


plexapi/utils.py
```diff
@@ -176,7 +176,11 @@ def toDatetime(value, format=None):
     """
     if value and value is not None:
         if format:
-            value = datetime.strptime(value, format)
+            try:
+                value = datetime.strptime(value, format)
+            except ValueError:
+                # parsing failed
```
Hellowlol: LOG.info('Fail to parse %s to datetime, defaulting to None', value)

makeroo: Yes, of course.
I'll submit a new PR.

Hellowlol:
Arh, sorry for the nitpicking, i should have explained better from the get go.
plexapi already have a logger that is configured. Can you reuse that one? 

```
from plexapi import log
```

makeroo:
Ok, I fixed it. I initialized LOG because grepping I found some modules doing the same, although I misread grep results: only flake8 and its submodules initialize the logger like that.

I'm sorry for the confusion, I did not analize plexapi logging configuration, that was the reason I put just a comment in the first place.

PR #368, opened 2019-07-27
Update library.py

h1f0x:
Issue when passing kwargs to url:

plexapi.exceptions.BadRequest: (400) bad_request; https://...31.plex.direct:32400/library/sections?name=6J0VMAS4&type=movie&agent=com.plexapp.agents.imdb&scanner=Plex+Movie+Scanner&language=de&location=%2FUser%2Fmoviescollections=False&extras=False&includeInGlobal=False

-> path is /user/movies
-> collections=False is kwarg

No '&' between.


h1f0x:
my bad.. not needed.

Issue #221, opened 2017-10-24
Support smart playlist


Hellowlol:
post 

https://ip:32400/playlists?type=video&title=TV%20Shows&smart=1&uri=library%3A%2F%2Fc6aaf29c-85a2-40dc-9a42-71ebcd615c9f%2Fdirectory%2F%252Flibrary%252Fsections%252F2%252Fall%253Ftype%253D4%2526episode.viewCount%25253E%25253E%253D0%2526sourceType%253D4&X-Plex-Product=Plex%20Web&X-Plex-Version=3.20.7&X-Plex-Client-Identifier=fzzf&X-Plex-Platform=Chrome&X-Plex-Platform-Version=61.0&X-Plex-Device=Windows&X-Plex-Device-Name=Plex%20Web%20%28Chrome%29&X-Plex-Device-Screen-Resolution=1280x615%2C1280x720&X-Plex-Token=xxx

Hellowlol:
@pkkid Any idea what travis is complaining about in my branch? https://github.com/pkkid/python-plexapi/tree/smart

pkkid:
I noticed sometimes things fail and if I kick it off again it will pass. I just started you're one failing build again, we can see if this is the case. -- We should figure it out and be a bit ore graceful, but I never put time into it yet.

blacktwin:
Some of the FILTERS or sectionChoices are not working properly when attempting to create a Smart Playlist.

`plex.createPlaylist("test", section="Movies", smart=True, unwatched=True, **{'genre':"Horror"})`
This will not user the unwatched FILTER/Choice, instead you just get the genre filter. Also because `title` is used for the naming of the Playlist `**{'title':'Die Hard'}` will return: `TypeError: create() got multiple values for argument 'title'`.  Maybe changing the parameter to `name` or `playlist_title`?

As it happens, looks like I accidentally omitted `title` from the `ALLOWED_FILTERS` for the MovieSection in PR #364 .

blacktwin:
PR incoming for `playlist.py`

```python
        for category, value in kwargs.items():
            if section.listChoices(category):
                sectionChoices = section.listChoices(category)
                for choice in sectionChoices:
                    if str(choice.title).lower() == str(value).lower():
                        uri = uri + '&%s=%s' % (category.lower(), str(choice.key))
            ####New check for BOOLEN FILTER
            if category in section.BOOLEAN_FILTERS:
                int_bool = section._cleanSearchFilter(category, value)
[...]

PR #370, opened 2019-07-31
allow for title and unwatched filters

blacktwin:
Addressing issue #221

### DIFF
--- a/plexapi/playlist.py
+++ b/plexapi/playlist.py
@@ -149,12 +149,12 @@ class Playlist(PlexPartialObject, Playable):
         return cls(server, data, initpath=key)
 
     @classmethod
-    def create(cls, server, title, items=None, section=None, limit=None, smart=False, **kwargs):
+    def create(cls, server, playlistTitle, items=None, section=None, limit=None, smart=False, **kwargs):
         """Create a playlist.
 
         Parameters:
             server (:class:`~plexapi.server.PlexServer`): Server your connected to.
-            title (str): Title of the playlist.
+            playlistTitle (str): Title of the playlist.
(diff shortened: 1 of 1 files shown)


pkkid:
It's small, but this is an API breaking change and needs the major version incremented before pushing to pypi.

blacktwin:
Sorry for being naive, but why, how? And what can I do to assist?

pkkid:
Not naive, just that a variable name change in a function definition will break existing code. If someone is calling that function and specifying `playlistTitle=something`, that will break if they update to whatever version of the code this change lands in. We when we update pypi, we just need to make sure we update the major version number so they know to expect these types of changes.

blacktwin:
@Hellowlol With the exception of aforementioned major version update, is there anything else I can do for this PR?

Hellowlol:
This dont seems to be a breaking change? He didnt change a keyword argument. Why was the title changed btw?

blacktwin:
> Why was the title changed btw?

The change would allow for the use of kwarg `title=` for filtering. 
```python
# Currently
plex.createPlaylist(title="Naming of Playlist", title="Title Filter") # will not work
plex.createPlaylist(title="Naming of Playlist", **{"title":"Title Filter"}) # will work
# PR
plex.createPlaylist(playlistTitle="Naming of Playlist", title="Title Filter")
plex.createPlaylist(playlistTitle="Naming of Playlist", **{"title":"Title Filter"})
```

Hellowlol:
Ah, i forgot about smart playlists. In the mean time lets just push this back to the next version. Its possible to handle  this correctly by using _createSmart

JonnyWong16:
Changed in #763.

PR #378, opened 2019-09-02
Fix for getting Episode by index from Season

menushka:
The Episode object doesn't contain the 'seasonNumber' attribute, resulting all queries to get an Episode from a Season using index to result in NotFound.  I replaced 'seasonNumber' with 'parentIndex', the proper attribute on the Episode object.


Hellowlol:
Nice! nice find. Can you write a test for this?

Hellowlol:
Thanks @menushka

Issue #356, opened 2019-06-30
Missing features


Hellowlol:
This is just a list of features that I have noticed that are not possible with plexapi. If you have time to help out, we would love if you choose one of these areas.

- [x] Upload subtitles
- [ ] New stats page

blacktwin:
For anyone who may want to start adding these.

POST Upload subtitles 
`http://127.0.0.1:32400/library/metadata/{RATING_KEY}/subtitles?title={SUB_FILE}.srt
&format=srt`

GET dashboard
`http://127.0.0.1:32400/statistics/bandwidth?timespan=6`

`http://127.0.0.1:32400/statistics/resources?timespan=6`

`http://127.0.0.1:32400/statistics/media?timespan=3`

`http://127.0.0.1:32400/library/all/top?type=2&viewedAt%3E=1565723660&limit=6`

JonnyWong16:
Both #384 and #636 have been merged.

PR #384, opened 2019-09-23
upload, select, remove subtitles

blacktwin:
Movies and Episodes can now have subtitle files uploaded, selected (for user?), and removed.
Episodes was missing subtitleStreams function.

Referencing #356

### DIFF
--- a/plexapi/video.py
+++ b/plexapi/video.py
@@ -3,7 +3,7 @@ from plexapi import media, utils
 from plexapi.exceptions import BadRequest, NotFound
 from plexapi.base import Playable, PlexPartialObject
 from plexapi.compat import quote_plus
-
+import ntpath
 
 class Video(PlexPartialObject):
     """ Base class for all video objects including :class:`~plexapi.video.Movie`,
@@ -226,6 +226,32 @@ class Movie(Playable, Video):
             for part in elem.parts:
                 streams += part.subtitleStreams()
         return streams
+
+    def uploadSubtitles(self, file):
+        """ Upload Subtitle file for video. """
+        url = '%s/subtitles' % self.key
+        filename = ntpath.basename(file)
+        with open(file, 'rb') as subfile:
+            params = {'title': filename,
+                      'format': filename.rsplit('.', 1)[1]
+                      }
+            headers = {'Accept': 'text/plain, */*'}
+            self._server.query(url, self._server._session.post, data=subfile, params=params, headers=headers)
+        self.reload()
+
+    def selectSubtitle(self, streamID=int, streamTitle=str):
+        url = '/library/parts/%s' % self.media[0].parts[0].id
+        for stream in self.subtitleStreams():
+            if streamID == stream.id or streamTitle == stream.title:
+                params = {'subtitleStreamID': stream.id}
+                self._server.query(url, self._server._session.put, params=params)
+        self.reload()
+
+    def removeSubtitles(self, streamID=int, streamTitle=str):
+        for stream in self.subtitleStreams():
+            if streamID == stream.id or streamTitle == stream.title:
+                self._server.query(stream.key, self._server._session.delete)
+        self.reload()
 
     def _prettyfilename(self):
         # This is just for compat.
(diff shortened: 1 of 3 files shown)


blacktwin:
Tests incoming.

plexapi/video.py
```diff
@@ -227,6 +227,32 @@ def subtitleStreams(self):
...
+        filename = ntpath.basename(file)
+        with open(file, 'rb') as subfile:
+            params = {'title': filename,
+                      'format': filename.rsplit('.', 1)[1]
+                      }
+            headers = {'Accept': 'text/plain, */*'}
+            self._server.query(url, self._server._session.post, data=subfile, params=params, headers=headers)
+        self.reload()
+
+    def selectSubtitle(self, streamID=int, streamTitle=str):
```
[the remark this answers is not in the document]
blacktwin: Looks like it. I guess I'll revert and commit without this function. Probably should have committed each function.

blacktwin:
Let me know if there is anything else messing or if I messed up the PR.

plexapi/video.py
```diff
@@ -622,3 +643,31 @@ def show(self):
     def _defaultSyncTitle(self):
         """ Returns str, default title for a new syncItem. """
         return '%s - %s - (%s) %s' % (self.grandparentTitle, self.parentTitle, self.seasonEpisode, self.title)
+
+    def subtitleStreams(self):
+        """ Returns a list of :class:`~plexapi.media.SubtitleStream` objects for all MediaParts. """
+        streams = []
+        for elem in self.media:
```
Hellowlol: use iterParts()

plexapi/video.py
```diff
@@ -622,3 +643,31 @@ def show(self):
...
+        subFormat = os.path.splitext(filepath)[1][1:]
+        with open(filepath, 'rb') as subfile:
+            params = {'title': filename,
+                      'format': subFormat
+                      }
+            headers = {'Accept': 'text/plain, */*'}
+            self._server.query(url, self._server._session.post, data=subfile, params=params, headers=headers)
+        self.reload()
+
+    def removeSubtitles(self, streamID=int, streamTitle=str):
```
Hellowlol: We dont use any typing so remove streamID=int, streamTitle=str abd just use None, you you want it to be a empty string for some reason use ''

Hellowlol: Also, does this delete the subtitle on the server or just disables it for this item?

tests/test_video.py
```diff
@@ -83,6 +83,29 @@ def test_video_Movie_download(monkeydownload, tmpdir, movie):
...
+def test_video_Movie_upload_select_remove_subtitle(movie, subtitle):
+    import os
+    filepath = os.path.realpath(subtitle.name)
+    movie.uploadSubtitles(filepath)
+    subtitles = [sub.title for sub in movie.subtitleStreams()]
+    subname = subtitle.name.rsplit('.', 1)[0]
+    assert subname in subtitles
+
+    subtitleSelection = movie.subtitleStreams()[0]
+    movie.selectSubtitle(streamID=subtitleSelection.id)
```
Hellowlol: You have removed this function so the test will fail.

plexapi/video.py
blacktwin: It depends on where the subtitles file is located. If it's inside the video's folder then it will be deleted. If outside of the video's folder then Plex will not remove it. I should probably add that in the docstring.

plexapi/video.py
blacktwin: This was a copy paste from the same function found in[ Movie class.](https://github.com/pkkid/python-plexapi/blob/master/plexapi/video.py#L222-L228). Want me to correct both? Or correct it once and place it in the Video class since both Movie and Episode load the Video class.

blacktwin: In the latest commit this function and the other subtitle related functions have been moved to the Video class.

plexapi/video.py
Hellowlol: Can you fix the doc string?

blacktwin: It's been [updated](https://github.com/blacktwin/python-plexapi/blob/patch-8/plexapi/video.py#L116-L120).

tests/test_video.py
blacktwin: This was also fixed.

Hellowlol:
Good stuff. Can you rebase so i can merge this. I’m not sure about the reloads. I don’t like reloading without the user explicitly does it. (For actions that is) this isn’t issue with this pr, but as a general issue.

blacktwin:
I can double check to see if the reload is necessary. You said rebase, but it looks like I just merged master in the PR. If this PR is too jacked up, I can close and create a clean PR.

Hellowlol:
Reload is only needed for the plexapi object updated (request the updated xml data from the server)

Issue #385, opened 2019-09-29
Media Part - Accessible / Exists / Available


animosity22:
I can see in the XML, the media part with the location for the file name will show if it's available or not available.

A nice user on reddit detailed what I was looking for and typed up what needed to be done to get access to the attribute in the API:

https://www.reddit.com/r/PleX/comments/davwpo/pythonplexapi_unavailable_status_for_a_location/

This was the message:

> I could be wrong, but from what I can see they python plex api currently doesn't have the "accessable" item under a Media objects Parts in the attributes as something that can be queried. You could request that be added. It If you want to add it yourself you can add
> self.accessible = data.attrib.get('accessible') 
> to the media.py file around line 107. Once that is added you can then get that information by looking at the parts of the movie object. something like
> video.reload()
> for m in video.media:
>     for p in m.parts:
>         print (p.accessible)
> If you look at the xml info you'll notice plex has it setup to have that information for each part of each media object. ie, if you have the multiple copies of a movie, and each copy is made up of multiple files(part 1, part 2...) than this will show you that information for each part for each movie.

This is the screenshot as well:

<img width="121" alt="image" src="https://user-images.githubusercontent.com/5088712/65837297-8f2b7200-e2c3-11e9-975a-29eb368e1e76.png">

blacktwin:
While the suggestion feels like it should work, it doesn't. For some reason the `accessible` and `exits` attribs are not getting pulled in with the object's xml `_data`.  Looks like `?checkFiles=1` needs to be added to the initial url string.

blacktwin:
This issue can be approached a couple of ways. 
Either edit the base.py to include the `?checkFiles=1` call and include `self.accessible` and `self.exists` to the media.py or only make the media.py edits and have users reload the video object with the `_details_key` in order to see the added exists and accessible values:

```python
movie = plex.fetchItem(107063)
movie = plex.fetchItem(movie._details_key)
print(movie.media[0].parts[0].exists)
```
And lastly, the solution I've provided in the referenced PR.

pkkid:
I don't have a good answer on which to choose. 

- I'm not against adding `?checkFiles=1` all the time. My only concern is we don't really know the impact of having it on all the time. Since Plex is a closed source app, we may never know.  Is it a slower call on the server side if we do that?
- If it is slower, I would fall back to making it an active choice to get the information.

animosity22:
[...]

PR #386, opened 2019-09-30
add exists function

blacktwin:
instead of modifying media.py and base.py, create a function to check if the video exists, is accessible, and which file if multiples exist(ed). For #385

### DIFF
--- a/plexapi/media.py
+++ b/plexapi/media.py
@@ -85,6 +85,8 @@ class MediaPart(PlexObject):
             key (str): Key used to access this media part (ex: /library/parts/46618/1389985872/file.avi).
             size (int): Size of this file in bytes (ex: 733884416).
             streams (list<:class:`~plexapi.media.MediaPartStream`>): List of streams in this media part.
+            exists (bool): Determine if file exists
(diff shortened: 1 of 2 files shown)


blacktwin:
First attempt at this issue. media.py and base.py can be updated for if exists. This was my minimal impact solution.

Hellowlol:
I would prefer if the we just add the missing attributes and require a manual reload.

blacktwin:
Yeah, I was thinking the same. What about the exists function for the video object?

blacktwin:
#385 exists and accessible will be cast to Boolean values. Video objects can pull this information along with which file the states are referring too with `video.exists()`. Not the prettiest return but I think it serves. Also added tests to validate the updated attributes.

Hellowlol:
Im not that fond of the exists function. I think it’s to general as you could end up with a part that is accessible and one that isn’t (2 files stored on different location) if a user want to check this they can use iterParts() and check what they want.

blacktwin:
Sounds good to me. I just figured it'd be a quicker solution to find that kind of information by accessing it at the video object vs media. It can be removed.

PR #387, opened 2019-10-05
Plexamp workaround for bad responses

jjlawren:
Plexamp responds to commands with `b'OK'` which is not valid XML, leading to a parser error on every response. This ignores that specific response.

### DIFF
--- a/plexapi/client.py
+++ b/plexapi/client.py
@@ -165,8 +165,6 @@ class PlexClient(PlexObject):
             log.warning('BadRequest (%s) %s %s; %s' % (response.status_code, codename, response.url, errtext))
             raise BadRequest('(%s) %s; %s %s' % (response.status_code, codename, response.url, errtext))
         data = response.text.encode('utf8')
-        if data == b'OK':   # Workaround for misbehaving clients such as Plexamp
-            data = ''
         return ElementTree.fromstring(data) if data.strip() else None
 
     def sendCommand(self, command, proxy=None, **params):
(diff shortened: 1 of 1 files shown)


jjlawren:
Okay, after a few iterations I think this is a better way. It still cannot inspect the data to ensure it's receiving `b'OK'`, but doing it at `sendCommand` is a better location.

This should also fix a bug where the PTP workaround in `sendCommand` wasn't respecting the proxy instance attribute.

plexapi/client.py
```diff
@@ -187,24 +187,27 @@ def sendCommand(self, command, proxy=None, **params):
...
-
-        if proxy:
-            return self._server.query(key, headers=headers)
-        return self.query(key, headers=headers)
+        try:
+            return query(key, headers=headers)
+        except ElementTree.ParseError:
+            # Workaround for players which don't return valid XML on successful commands
+            #   - Plexamp: `b'OK'`
+            if self.product in ('Plexamp'):
```
Hellowlol: Shouldnt this be ('Plexamp',) <-- note the comma

plexapi/client.py
```diff
@@ -187,24 +187,27 @@ def sendCommand(self, command, proxy=None, **params):
...
+                return None
```
Hellowlol: return is enough.

plexapi/client.py
jjlawren: Yes, it should. Fixed.

Issue #179, opened 2017-08-21
Support Changing Posters


JonnyWong16:
Someone found the endpoints to change poster images outlined in this Reddit post. I haven't tested this for myself yet.

https://www.reddit.com/r/PleX/comments/6ult18/any_way_to_automatically_cycle_posters_for_tv_and/dlw4awm/

Upload image:
```
POST https://localhost:32400/library/metadata/24633/posters?url=HTTP_URL_ENCODED_PATH&X-Plex-Token=PLEX_TOKEN
```

Then immediately GET to get the updated list of art options:
```
GET https://localhost:32400/library/metadata/24633/posters?X-Plex-Token=PLEX_TOKEN
```

Which gives an XML Response:

```
<?xml version="1.0" encoding="UTF-8"?>
<MediaContainer size="4" identifier="com.plexapp.plugins.library" mediaTagPrefix="/system/bundle/media/flags/" mediaTagVersion="1501029046">
    <Photo key="/library/metadata/24633/file?url=upload%3A%2F%2Fposters%2Fseasons%2F1%2Fepisodes%2F8%2F503c50f27618447703442227788980ac0c9d8610" ratingKey="upload://posters/seasons/1/episodes/8/503c50f27618447703442227788980ac0c9d8610" thumb="/library/metadata/24633/file?url=upload%3A%2F%2Fposters%2Fseasons%2F1%2Fepisodes%2F8%2F503c50f27618447703442227788980ac0c9d8610" selected="1" provider="" />
    <Photo key="/library/metadata/24633/file?url=metadata%3A%2F%2Fseasons%2F1%2Fepisodes%2F8%2Fthumbs%2Fcom%2Eplexapp%2Eagents%2Elocalmedia_92206304aa517b8756d458b05f3ffaebda9d044f" ratingKey="metadata://seasons/1/episodes/8/thumbs/com.plexapp.agents.localmedia_92206304aa517b8756d458b05f3ffaebda9d044f" thumb="/library/metadata/24633/file?url=metadata%3A%2F%2Fseasons%2F1%2Fepisodes%2F8%2Fthumbs%2Fcom%2Eplexapp%2Eagents%2Elocalmedia_92206304aa517b8756d458b05f3ffaebda9d044f" selected="0" provider="com.plexapp.agents.localmedia" />
[...]

PR #390, opened 2019-10-09
posters

blacktwin:
Initial work for #179. Adding poster objects first to more easily address issue #179. This PR can be a standlone as the rest of the request in #179 will be coming in a sperate PR.

### DIFF
--- a/plexapi/video.py
+++ b/plexapi/video.py
@@ -24,6 +24,7 @@ class Video(PlexPartialObject):
             type (str): 'artist', 'album', or 'track'.
             updatedAt (datatime): Datetime this item was updated.
             viewCount (int): Count of times this item was accessed.
+            posters (List<:class:`~plexapi.media.Poster`>): List of poster objects.
     """
 
     def _loadData(self, data):
@@ -42,6 +43,7 @@ class Video(PlexPartialObject):
         self.type = data.attrib.get('type')
         self.updatedAt = utils.toDatetime(data.attrib.get('updatedAt'))
         self.viewCount = utils.cast(int, data.attrib.get('viewCount', 0))
+        self.posters = self._posters()
 
     @property
     def isWatched(self):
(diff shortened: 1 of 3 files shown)


plexapi/video.py
```diff
@@ -89,6 +91,14 @@ def _defaultSyncTitle(self):
         """ Returns str, default title for a new syncItem. """
         return self.title
 
+    def _posters(self):
```
Hellowlol: Just make this a normal function no _

plexapi/video.py
```diff
@@ -42,6 +43,7 @@ def _loadData(self, data):
         self.type = data.attrib.get('type')
         self.updatedAt = utils.toDatetime(data.attrib.get('updatedAt'))
         self.viewCount = utils.cast(int, data.attrib.get('viewCount', 0))
+        self.posters = self._posters()
```
Hellowlol: Delete this

plexapi/video.py
```diff
@@ -24,6 +24,7 @@ class Video(PlexPartialObject):
             type (str): 'artist', 'album', or 'track'.
             updatedAt (datatime): Datetime this item was updated.
             viewCount (int): Count of times this item was accessed.
+            posters (List<:class:`~plexapi.media.Poster`>): List of poster objects.
```
Hellowlol: Remove this.

tests/test_video.py
```diff
@@ -101,6 +101,7 @@ def test_video_Movie_attrs(movies):
     assert [i.tag for i in movie.directors] == ['Nina Paley']
     assert movie.duration >= 160000
     assert movie.fields == []
```
Hellowlol: Update the test.

Hellowlol:
If it does a http call it should be a function or do the http when needed. With the current pr we will do a http call on every item plexapi builds.

blacktwin:
So leave it as a function and not a attribute. 👍  Can do. Thanks!

plexapi/video.py
```diff
@@ -89,6 +89,14 @@ def _defaultSyncTitle(self):
         """ Returns str, default title for a new syncItem. """
         return self.title
 
+    def posters(self):
+        """ Returns list of available poster objects. :class:`~plexapi.media.Poster`:"""
+        items = []
```
Hellowlol: I think you can do ```return self.fetchItems('%s/posters', cls=media.Poster)```

blacktwin: It doesn't appear like we can. The xml for posters does not have a `filter` or `tag` to be able to use the `fetchItems`. Edit above results in a empty list.

Hellowlol: The for this item should be Photo, that the tag i get when i try to do this. Can change the tag and try again?

Hellowlol: So it worked?

blacktwin: Yup. Not 100% sure why, but it does and you seem to know how/why.

Hellowlol:
Good stuff? Can you make a PR for art also?

Issue #397, opened 2019-11-06
Accessing some attributes can wipe valid objects


jjlawren:
It seems that sometimes when accessing certain attributes, it can "reset" the parent object.

I found that when accessing the `session` attribute on one of the `plexserver.sessions()` objects, it would occasionally clear the whole parent. The Plex server sometimes doesn't include the `Session` key in the XML when calling `/status/sessions`, and it's wiped out by this magic: 
https://github.com/pkkid/python-plexapi/blob/master/plexapi/base.py#L277-L291 

Here's an example to reproduce: https://gist.github.com/jjlawren/abbde9585ff25bece264ab3fc7dc8dbc

During the second iteration, it will wipe out all of the useful data _if_ the `Sessions` key is missing.

I don't know if other attributes can cause this to happen.

pkkid:
Do you have any clues what might may cause the session key to be missing? -- I ran the two for loops on my machine with a playing tvshow and movie and am getting expected results; but the sessions key is not missing for me.

jjlawren:
No, it's not readily reproducible, but I've definitely seen it when playing music most often. 

I've encountered the issue when playing an album, played a movie instead, and then went back to the same music track and the missing `Session` key appeared to clear itself up. The same Plex client was used through all those steps.

pkkid:
I'm able to force the bug to happen. -- When we see None or empty list, we assume the object returned from Plex is incomplete and reload the official full object from the url defined by `self.key`.  The problem here is that new object doesn't store any information about the playing sessions of that object, and they get overwritten with the empty data in `Playable()`.

A few small choices in code changes I can think of to help you get by this and still maintain reloading objects for the more common scenarios:
1. We add a specific check in  __getattribute__ similar to [line 281](https://github.com/pkkid/python-plexapi/blob/master/plexapi/base.py#L281) specifically to skip reload for the 'session' attribute.
2. We can add a class variable DONT_RELOAD_FOR_KEYS = ['key', 'session'] to the PlexPartialObject class, and use that. This would be OK because it makes it somewhat configurable the next time we hit it.
3. We don't overwrite values on any objects that are missing in the reloaded object. -- I think this option is a very large code change, and needs to be thought out a lot more. In some cases updating a class variable to None for missing information may actually be the correct thing to do.

I like option #2 above because the change is minimal and the use case seems small.

pkkid:
Try replacing the top section of `PlexPartialObject` with the following..

```python
[...]

PR #398, opened 2019-11-10
Skip attr reload for 'session'

jjlawren:
Fix as suggested by @pkkid in #397.

As described there:
> When we see None or empty list, we assume the object returned from Plex is incomplete and reload the official full object from the url defined by self.key. The problem here is that new object doesn't store any information about the playing sessions of that object, and they get overwritten with the empty data in Playable().


pkkid:
Thanks.

PR #404, opened 2019-11-18
Speed up Library Section Get

zSeriesGuy:
The original code pulled a full list of all library items on a library.section.get(title) and then searched that list.
With this change, the library get uses the title as a search filter such that the returned list is much shorter.
My tests show a significant performance improvement.

### DIFF
--- a/plexapi/library.py
+++ b/plexapi/library.py
@@ -374,7 +374,7 @@ class LibrarySection(PlexObject):
             Parameters:
                 title (str): Title of the item to return.
         """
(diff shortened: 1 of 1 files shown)


Hellowlol:
It much faster but I dont think this is the correct fix as the title is injected in the key. I dont think plexapi quotes anything later on. If you need something fast the user should probably search for something or handles quotes on the pr.

zSeriesGuy:
I traced thru the request. The spaces in the title do get url encoded and the spaced changed to %20.

Issue #401, opened 2019-11-12
required websocket-client==0.48.0 is old


onkelbeh:
Hi,

Could you please update websocket-client==0.48.0 in requirements.txt.
0.48.0 one is from May 2018. There are 8 newer version available.

Home Assistant requires websocket-client==0.54.0.

Thanks in advance.

Hellowlol:
Its old as the websocket-client had issues where the bug got added removed then added. Imo we should drop the entire library and use another one. (or atleast remove it from requirements.txt as its only needed for a small part.)

onkelbeh:
Thanks, can not tell if 0.54.0 is buggy or not, just have lots of installation problems with hard run dependencies, especially for often used libraries (when they get older). I use it in my [Home Assistant Overlay for Gentoo Linux](https://github.com/onkelbeh/HomeAssistantRepository), where i have to take care of _some_ other dependencies, too ;-)

zSeriesGuy:
I saw in issue #353 that there was an issue with the AlertListener callback unless websocket-client is pinned at 0.47 or 0.53.
I have been testing using 0.56 (latest release), and the "not getting called" issue is back, starting with 0.54. It raises an exception, something about missing 1 required positional parameter, message.

If I remove the "ws" from AlertListener._onMessage method def, the alert callbacks work correctly with the latest release of websocket-client. Probably needs to be removed from _onError also.

I don't know if this is an acceptable solution. But wanted to let you know in case you want to get past the pinning of websocket-client.

Hellowlol:
Seems fine by me.

PR #408, opened 2019-12-04
Update websockets client

zSeriesGuy:
Looks like websockets-client returns 2 args in 0.48.0 while in the latest release, it only returns 1 arg.
So I used *args to handle the difference in releases and assume that the last arg is the message.

This should resolve issue #401

### DIFF
--- a/plexapi/alert.py
+++ b/plexapi/alert.py
@@ -6,9 +6,9 @@ from plexapi import log
 
 
 class AlertListener(threading.Thread):
-    """ Creates a websocket connection to the PlexServer to optionally recieve alert notifications.
+    """ Creates a websocket connection to the PlexServer to optionally receive alert notifications.
         These often include messages from Plex about media scans as well as updates to currently running
-        Transcode Sessions. This class implements threading.Thread, therfore to start monitoring
+        Transcode Sessions. This class implements threading.Thread, therefore to start monitoring
         alerts you must call .start() on the object once it's created. When calling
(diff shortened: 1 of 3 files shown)


plexapi/alert.py
```diff
@@ -55,7 +55,8 @@ def stop(self):
         log.info('Stopping AlertListener.')
         self._ws.close()
 
-    def _onMessage(self, ws, message):
+    def _onMessage(self, *args):
+        message = args[-1]
```
pkkid: Can you please add a comment here explaining the wonkyness between versions and why we're grabbing the last item? We're all going to forget in about 3 weeks time. :D

plexapi/alert.py
```diff
@@ -65,6 +66,7 @@ def _onMessage(self, ws, message):
         except Exception as err:  # pragma: no cover
             log.error('AlertListener Msg Error: %s', err)
 
-    def _onError(self, ws, err):  # pragma: no cover
+    def _onError(self, *args):  # pragma: no cover
+        err = args[-1]
```
pkkid: Same, lets comment this.

zSeriesGuy: Good point. Done.

pkkid:
Thanks, looks good to me. Going to let others review before merging.

Hellowlol:
I agree :)

pkkid:
Thanks for the fix! :D

PR #410, opened 2019-12-05
new_hubs

blacktwin:
Allow for pulling information from hub sections
Disclaimer, Token will need to be pulled from web browser session using app.plex.tv.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,8 +10,6 @@ services:
   - docker
 
 python:
-  - 2.7
-  - 3.4
   - 3.6
 
 env:
@@ -23,8 +21,10 @@ env:
 before_install:
   - pip install --upgrade pip
   - pip install --upgrade setuptools
-  - pip install --upgrade pytest pytest-cov coveralls
 install:
+  - pip install .
+  - python -c "import plexapi; print('installation ok')"
+  - pip install --upgrade pytest pytest-cov coveralls
   - pip install -r requirements_dev.txt
   - '[ -z "${PLEXAPI_AUTH_MYPLEX_USERNAME}" ] && PYTHONPATH="$PWD:$PYTHONPATH" python -u tools/plex-bootstraptest.py
      --destination plex --advertise-ip=127.0.0.1 --bootstrap-timeout 540 --docker-tag $PLEX_CONTAINER_TAG --unclaimed ||
--- a/plexapi/alert.py
+++ b/plexapi/alert.py
@@ -6,9 +6,9 @@ from plexapi import log
 
 
 class AlertListener(threading.Thread):
-    """ Creates a websocket connection to the PlexServer to optionally recieve alert notifications.
+    """ Creates a websocket connection to the PlexServer to optionally receive alert notifications.
         These often include messages from Plex about media scans as well as updates to currently running
-        Transcode Sessions. This class implements threading.Thread, therfore to start monitoring
+        Transcode Sessions. This class implements threading.Thread, therefore to start monitoring
         alerts you must call .start() on the object once it's created. When calling
         `PlexServer.startAlertListener()`, the thread will be started for you.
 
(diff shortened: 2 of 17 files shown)


pkkid:
Wow you're fast! Is this already ready to go?

blacktwin:
Kinda. I'm not 100% sure about where I put everything. Would appreciate any feedback. It pulls the information fine but I think there is more to add as well.

blacktwin:
Much faster in returning the maximum amount of Hubs within the providers.

blacktwin:
I'd like to find a way to get [this](https://old.reddit.com/r/PleX/comments/e6ocvc/a_complete_list_of_movies_available_from_plex/?st=k3tlup0n&sh=d7fdc3c5) in there too.

mcqweb:
> I'd like to find a way to get [this](https://old.reddit.com/r/PleX/comments/e6ocvc/a_complete_list_of_movies_available_from_plex/?st=k3tlup0n&sh=d7fdc3c5) in there too.

Just added the paging to that too now so it genuinely should be the full list (of unwatched anyway)

blacktwin:
I think this is good for now. The ability to query all available free items is available if you want and the additions return what Plex returns.

blacktwin:
I'm stomped. The new endpoint defined in myplex.py will return the xml with 
```python
requests.get(VOD, params={'X-Plex-Token':TOKEN})
```
but not when using the methods added to this PR. Any help would be very appreciated.

Hellowlol:
All logging and simplify. Are you use the end point support the plex x container param etc. use fetch items directly

blacktwin:
> Are you use the end point support the plex x container param etc 

Yes.

>is the token required to be in the url? 

Yes.

I'll looking into logging and see what I get.

Hellowlol:
The token is probably the reason why it don’t work. The token is added in the header not the url (see the class.query) Some endpoint require different stuff

blacktwin:
This works:
```python
    def webShows(self, maxresults=50):
        """ Returns a list of Webshow Hub items :class:`~plexapi.library.Hub`
		"""
        req = requests.get(self.WEBSHOWS, headers={'X-Plex-Token':self._token})
        elem = ElementTree.fromstring(req.text)
        return self.findItems(elem)
```
but not:
```python
self.query(self.WEBSHOWS)
```
Which was the method I used [before ](bd2519b35b6c2a6b133d20b5fe7d45c6dcbd70a4)and it definitely did work. Very odd.

So I guess I'll go with the working method above.

blacktwin:
From digging into these endpoints more we may want to build out another section for `providers`. Which are outside media items sources.
```python
    VOD = 'https://vod.provider.plex.tv/hubs/'
    WEBSHOWS = 'https://webshows.provider.plex.tv/hubs/'
    NEWS = 'https://news.provider.plex.tv/hubs/'
    PODCASTS = 'https://podcasts.provider.plex.tv/hubs/'
    MUSIC = 'https://music.provider.plex.tv/hubs/'
    METADATA = 'https://metadata.provider.plex.tv/'  
```
`METADATA` being the endpoint to pull additional information from these outside sources. [Example](https://old.reddit.com/r/PleX/comments/edm420/api_for_the_newish_songs_in_this_episode_feature/fghpf2y/). 

Each provider endpoint can branch off from it initial root `*.provider.plex.tv` using the `key` tags in the xml. Example being `/hubs/sections/home/popular` would show the popular podcasts/news/etc. Selecting a single items from that list and using it's `key` tag and appending it to the `METADATA` endpoint would provide all the metadata for that item.

blacktwin:
Now everything is pulling basic Hub information. I think this is a good start but there is definitely more that can be added as far as the sub Items inside each Hub endpoint as mentioned above. I think that can be left for another PR as I think it will require a lot more changes. 

Let me know if there needs any changes, suggestions, tests, etc. I'd like to close this to start working on other issues/requests and come back to adding more of the sub Item work.

blacktwin:
Ready for review @Hellowlol

PR #411, opened 2019-12-13
Add support for MyPlex PIN login

Montellese:
I've added a class `myplex.MyPlexPinLogin` which supports PIN login using https://plex.tv/link which is mostly useful on clients using a 10-ft interface.

When creating a new instance of `myplex.MyPlexPinLogin` it asks Plex for a PIN and then it checks whether an authenticated user has entered the PIN. If that is the case it returns an access token which can be used to create a `myplex.MyPlexAccount` instance.

There are three modes:
* The **polling** approach expects the developer to periodically check if the PIN login was successful using `myplex.MyPlexPinLogin.checkLogin()`.
* The **threaded** approach expects the developer to call `myplex.MyPlexPinLogin.run()` and then at a later time call `myplex.MyPlexPinLogin.waitForLogin()` to wait for and check the result.
* The **callback** approach is an extension of the threaded approach and expects the developer to pass the `callback` parameter to the call to `myplex.MyPlexPinLogin.run`. The callback will be called when the thread waiting for the PIN login to succeed either finishes or expires. The parameter passed to the callback is the received authentication token or `None` if the login expired.

**DISCLAIMER**: I'm not a very experienced Python developer. I tried to follow the code style in `plexapi.myplex` but I'm sure I got some things wrong (especially in the docstrings).

### DIFF
--- a/plexapi/myplex.py
+++ b/plexapi/myplex.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 import copy
+import threading
 import time
 
 import requests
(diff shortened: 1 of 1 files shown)


pkkid:
This is a cool, thanks. the class essentially returns a token, correct? Would it make sense to call _connect instead so we can always return a PlexServer object instead? I don't really see a use case where we wouldn't want to do that.

Montellese:
Yes it first provides a `pin` that needs to be displayed to the user. And then once authentication has succeeded it provides a `token`. But don't you mean a `MyPlexAccount` instance because PIN login links a client to a MyPlex account and not to a specific server.

pkkid:
Ahh. I get it, thanks.

Montellese:
How / Where should I return the `MyPlexAccount` instance? Directly from `myplex.MyPlexPinLogin.checkLogin()` / `myplex.MyPlexPinLogin.waitForLogin()` (instead of `True` / `False`) or should I provide a separate method?

Montellese:
I have fixed an issue with manually calling `MyPlexPinLogin.checkLogin()` because it didn't properly set the `finished` and `expired` members.

@pkkid any pointers on how best to return a `MyPlexAccount` from `MyPlexPinLogin`?

pkkid:
I'd have to play with it a bit to get back into it. I haven't had much time to play recently.

Hellowlol:
Sorry for leaving this for so long. I dont think we need to return a account. The user can use the callback to return a account 

```

def cb(token)
    return MyPlexAccount(token=token)
```

I'll rebase and do a review of this. Just glanced at it and it seems ok. (i also need to write some tests)

Montellese:
Thanks for the feedback. I've rebased it and fixed the open conflict.

plexapi/myplex.py
```diff
@@ -1052,6 +1053,187 @@ def syncItems(self):
...
+        self._loginTimeout = None
+        self._callback = None
+        self._thread = None
+        self._abort = False
+        self._id = None
+
+        self.finished = False
+        self.expired = False
+        self.token = None
+        self.pin = None
```
Hellowlol, line 1106: Mind removing this. Pretty much the only thing misssing

Montellese: Nice spotting. Done and rebased.

PR #436, opened 2020-03-04
unmatch_match

blacktwin:
Adding methods to unmatch, find matches, and use found matches to Fix match movies and shows.


Hellowlol:
Cool!

blacktwin:
The key/url creation is kinda sloppy. I'll have to think about improvements. Now users can use the shorten agent names `themoviedb` instead of `com.plex.agent.themoviedb` or whatever it is for the agent field. 

An example of applying a new match:
```python
newMatches = movie.matches(auto=False, title='Good boy', agent='themoviedb')
for result in newMatches:
    if result.score == 100:
        movie.fixMatch(result)
        break
```

Hellowlol:
The SearchResult class, when is created? Is this just for the search result relating to the matches or can this creep up other places too?

blacktwin:
Yes, just for search results from looking for metadata from agents. I'm not really sure where it would creep into/from.

blacktwin:
update to library's `edit` method now allows for easier editing of the library settings.
example:
```python
enableBIF = plex.library.section("Movies").settings()[2].id
test = {'prefs[%s]' % enableBIF: 1}
plex.library.section("Movies").edit(**test)
```

blacktwin:
May have to revert those last two commits 2ab3b61, 98226c0 as they break the AgentMEdiaType

blacktwin:
Ready for review @Hellowlol . This PR allows for unmatching and searching, selecting, and matching metadata results from library agents. Updates to the library adding library agent and library settings.

PR #437, opened 2020-03-12
poster_change

blacktwin:
Allow for selecting a poster from the available list of posters. Allow for uploading posters from url or filepath.
From issue #179
Poster selection currently causes a parser error
```
xml.etree.ElementTree.ParseError: not well-formed (invalid token): line 1, column 0
```
but the poster will still be selected.

### DIFF
--- a/plexapi/base.py
+++ b/plexapi/base.py
@@ -445,6 +445,7 @@ class PlexPartialObject(PlexObject):
             self._server.query(key, method=self._server._session.post, data=data)
 
     def setPoster(self, poster):
+        """ Set . :class:`~plexapi.media.Poster` to :class:`~plexapi.video.Video` """
         key = poster._initpath[:-1]
         data = '%s?url=%s' % (key, quote_plus(poster.ratingKey))
         self._server.query(data, method=self._server._session.put)
--- a/plexapi/base.py
+++ b/plexapi/base.py
@@ -429,6 +429,26 @@ class PlexPartialObject(PlexObject):
         """
         return self._server.history(maxresults=maxresults, mindate=mindate, ratingKey=self.ratingKey)
 
+    def posters(self):
+        """ Returns list of available poster objects. :class:`~plexapi.media.Poster`. """
+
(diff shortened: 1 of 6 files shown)


Hellowlol:
If plex returns junk xml but the request still succeeds, just catch and swallow the exception.

Hellowlol:
Is the poster option only available for video files or is it possible to do this on albums etc?

blacktwin:
> If plex returns junk xml but the request still succeeds, just catch and swallow the exception.

it returns the new thumb path. `/library/metadata/80584/thumb/1584129784` So I'd assume make the changes in the server.py?

> Is the poster option only available for video files or is it possible to do this on albums etc?

I'll check. If so I'll add the method to the appropriate class(s).

blacktwin:
Moved everything to base.py. Now artists, albums, videos, movies, shows, etc should be covered.

Hellowlol:
Maybe you need to create another class to add your methods to. The plexpartial class uses several places. Can you doble check that we dont have any unwanted side effects? (Like the method is available somewhere it shouldn’t.)

blacktwin:
> If plex returns junk xml but the request still succeeds, just catch and swallow the exception.

My solution so far is to update the server.py `query` method's return from:
`return ElementTree.fromstring(data) if data.strip() else None`
to 
`return ElementTree.fromstring(data) if data.strip() and response.encoding else None`
The poster selection method returns `response.encoding = None` so this is one why of silencing the error.

blacktwin:
> Maybe you need to create another class to add your methods to. The plexpartial class uses several places. Can you doble check that we dont have any unwanted side effects? (Like the method is available somewhere it shouldn’t.)

Looks to only be used in location where posters would exist. `Playlist` maybe not, but IIRC you can hack the poster of a `Playlist`. 
image

blacktwin:
Also to note with changing the poster in Playlists, the poster you change to is also the faded background art.

blacktwin:
Ready for review @Hellowlol . This PR will allow for updating posters and art.

PR #443, opened 2020-03-30
Github action to publish to PyPI

jjlawren:
This is a Github Actions workflow to automatically publish tagged releases to PyPI as described [here](https://help.github.com/en/actions/language-and-framework-guides/using-python-with-github-actions#publishing-to-package-registries).

This will require @pkkid to:

1. Create a new token on PyPI under "Account Settings > API tokens"
2. Set it as `PYPI_TOKEN` in the repository's encrypted secrets

More details about the encrypted secrets here: https://help.github.com/en/actions/configuring-and-managing-workflows/creating-and-storing-encrypted-secrets.


pkkid:
Thank you for doing the research here and I'm sorry I have not had time to dedicate here. I'll take a look at these tonight.

pkkid:
The steps from the top comment are complete.

jjlawren:
I think this action should work independently. If it's a question regarding the state of the code & outstanding PRs before a new release, I'd have to defer back to you.

jjlawren:
Anything I can help with here @pkkid?

pkkid:
I'm sorry I misinterpreted the comment above, thinking I didn't need to do anything more here yet. 

I believe the next steps are to setup a file per these instructions: https://packaging.python.org/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/

Do we want it to push every time we do a tagged commit (per the example)?

jjlawren:
That's a different process for a similar result. The link above also doesn't seem to handle actually building the package, just publishing. Note from that page: https://github.com/pypa/gh-action-pypi-publish#non-goals.

I reviewed this PR and changed the trigger to only run the Github Action on published releases. So the workflow should be as simple as:
1) Ensure the token (not username/pass) is stored in this repo's secrets
2) Commit this Github Action
3) Publish a new Github release for the project

Once that's done the action should fire, spin up an Ubuntu instance, build the package, and upload it to PyPI using the token.

One thing that I _didn't_ consider is Python 2.x support. This action is only set to build for 3.x and I don't think there's been an explicit announcement of the project's end of support for 2.x. Thoughts?

Hellowlol:
Yes, since  you are the only one access to the pypi account it would be nice to be able to push a release, some projects cant install from github etc.

pkkid:
@Hellowlol Just added you on PyPi.

jjlawren:
> Yes, since  you are the only one access to the pypi account it would be nice to be able to push a release, some projects cant install from github etc.

The point of this script is that anyone that can publish a release on github (i.e., collaborators) can also push to PyPI.

I'd think that basing off releases instead of tags will make it easier to keep things in sync. For example, the "missing" 3.3.0 github release.

jjlawren:
@pkkid @Hellowlol if the PyPI token is in place, are there any objections to merging this PR and creating a 3.4.0 release from the current master?

Hellowlol:
I think we should create builds for Python 2 (3.4.0) then merge any outstanding prs, drop py2 and push 4.0.0

Hellowlol:
@pkkid yes, it should push a new version on each tag. We have support for this now using travis. But its only released if all tests pass (our tests are to flaky)

jjlawren:
Rebased with all the recent test fixes and it passes. Thanks, @Hellowlol!

Since Travis is already capable of pushing to PyPI (if all tests pass), what's the consensus on the necessity of this PR now?

JonnyWong16:
Updated workflow here:
https://github.com/JonnyWong16/python-plexapi/actions/runs/399032800/workflow
https://test.pypi.org/project/PlexAPI/

JonnyWong16:
Change `python_requires='>=3.6' in `setup.py`.

https://github.com/pkkid/python-plexapi/blob/master/setup.py#L38

jjlawren:
> Change `python_requires='>=3.6'` in `setup.py`.
> 
> https://github.com/pkkid/python-plexapi/blob/master/setup.py#L38

Let's do this at the next major release unless there's a reason it has to be done with this PR.

Hellowlol:
Travis is gone. Time to merge this?

PR #445, opened 2020-04-01
GDM server and client discovery

jjlawren:
Plex uses a proprietary protocol called GDM (Good Day Mate) which allows clients and servers to discover each other. It uses UDP sent to broadcast/multicast addresses, so it will only work when devices are on the same broadcast subnet. This limitation can be addressed by using a UDP repeater utility to forward UDP traffic between networks.

This module is based on an implementation originally from https://github.com/home-assistant/netdisco/netdisco/gdm.py.

I've also added support for Plex client discovery to the original implementation.

### DIFF
--- a/plexapi/gdm.py
+++ b/plexapi/gdm.py
@@ -11,14 +11,13 @@ Inspired by
 """
 import socket
 import struct
-from typing import Any, Dict, List  # noqa: F401
 
 
 class GDM:
     """Base class to discover GDM services."""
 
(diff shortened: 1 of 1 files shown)


Hellowlol:
Remove the typing or make them python2 compat. We plexapi is still compat with python 2. (We should remove py 2 support but that’s another story)

jjlawren:
That's done.

PR #446, opened 2020-04-02
Fix playMedia() calls with a PlayQueue

jjlawren:
I was getting exceptions when calling `PlexClient.playMedia()` with a `PlayQueue`:
```
  File "/usr/local/lib/python3.7/site-packages/plexapi/client.py", line 472, in playMedia
    elif media.listType == "audio":
AttributeError: 'PlayQueue' object has no attribute 'listType'
```
This gets the media type of the first item in the playqueue. Not perfect if someone has a playqueue mixed with music and video, but I don't see a better alternative.

### DIFF
--- a/plexapi/client.py
+++ b/plexapi/client.py
@@ -472,10 +472,12 @@ class PlexClient(PlexObject):
 
         if hasattr(media, "playlistType"):
             mediatype = media.playlistType
-        elif media.listType == "audio":
-            mediatype = "music"
         else:
-            mediatype = "video"
+            if isinstance(media, PlayQueue):
+                listType = media.items[0].listType
+            else:
+                listType = media.listType
+            mediatype = "music" if listType == "audio" else "video"
 
         if self.product != 'OpenPHT':
             try:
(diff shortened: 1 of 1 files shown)


jjlawren:
@Hellowlol this is related to your client fix.

Hellowlol:
Is it possible to mix video and music in a playqueue?

jjlawren:
Honestly it's not something I've tested as it's not something I have a need for. The only scenario where I could see that being useful is mixing music tracks + music videos.

I'd like to figure it out as this PR is required to use the current codebase in Home Assistant as it currently relies on passing in playqueues for playback. I'll try to find time to run some tests later today.

jjlawren:
Alright, so it seems that `PlayQueue.create()` currently only supports a single item or playlist passed to it. Playlists are enforced as either music or video by Plex itself (and also by `Playlist.create()`). I think it's safe to assume all items will be a similar mediatype here.

jjlawren:
Tested with movies, shows, albums, tracks, and playlists both directly and inside of playqueues. All seems to work properly now.

PR #447, opened 2020-04-08
Use Unauthorized exception

jjlawren:
The `Unauthorized` exception was declared but never used. This will raise the exception when a 401 error is encountered.

The exception has been changed to inherit from `BadRequest` to preserve existing use cases that catch on `BadRequest`. Same with the error message in case there's existing reliance on parsing.


jjlawren:
I should add this to at least myplex, too.

PR #451, opened 2020-04-15
Raise NotFound on 404 errors

jjlawren:
Raise a more appropriate `NotFound` exception instead of a `BadRequest` when encountering a 404 error.


Hellowlol:
Can you add some test?

jjlawren:
I added a test for the `server.py` exception. I haven't found a way to reproduce on the other classes in a real environment, but I also don't see a downside in case those can throw 404 errors. It also keeps the `query()` implementations similar for each.

PR #452, opened 2020-04-16
Fix docstrings for tests

jjlawren:
Minor fixes to allow documentation generation to pass tests.

Found and removed an unused param in `PlexPartialObject.matches()`.

### DIFF
--- a/plexapi/media.py
+++ b/plexapi/media.py
@@ -456,7 +456,7 @@ class Conversion(PlexObject):
                     Move 5th conversion Item to active conversion
                         conversions[4].move('-1')
 
-                    Move 4th conversion Item to 2nd in conversion queue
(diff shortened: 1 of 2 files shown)


plexapi/media.py
```diff
@@ -450,10 +450,7 @@ def remove(self):
 
     def move(self, after):
         """ Move Conversion items position in queue
-            after (int): Positional integer to move item
-                    -1 Active conversion
-                OR
-                    Use another conversion items playQueueItemID to move in front of
+            after (int): Place item after specified playQueueItemID. '-1' is the active conversion.
```
jjlawren, line 453: @blacktwin is this still accurate?

Hellowlol:
I think we should consider to change this method so we can use normal python index to move the items.

PR #459, opened 2020-04-17
Bump to 3.4.0

jjlawren:


pkkid:
This didn't make it to PyPi. Do we need to tag it or do some Github magic?

jjlawren:
Maybe because I added the release/tag after committing the PR. I'm trying a manual Travis build now: https://travis-ci.org/github/pkkid/python-plexapi/builds/676296441.

jjlawren:
The travis job reported:
> `stages`: skipping stage #`2` because its condition does not match: `tag = present`

I'm guessing we should do one of two things:

1. Change the stage condition from `tag = present` to `tag IS present` (https://docs.travis-ci.com/user/conditions-v1)
2. Remove the stage condition since the `deploy` step is already condition-based with `on: tags: true`. (https://docs.travis-ci.com/user/deployment#conditional-releases-with-on)

Hellowlol:
I think we should remove 1 and just let the deploy handle it

jjlawren:
Updated in #460. It might need a fresh tag to trigger once that's merged.

PR #460, opened 2020-04-17
Remove redundant travis conditional

jjlawren:


jjlawren:
@pkkid @Hellowlol: I _think_ the syntax was incorrect before and this should allow us to publish off the next tag. If either of you are able to upload a manual PyPI package for 3.4.0 to avoid tagging a redundant release it would be greatly appreciated.

PR #466, opened 2020-04-26
Make it easier to set a container size using search

Hellowlol:
~~Now we still 1 more http call then we need as we just asks for any data until the subresult is empty.~~

~~If we override fetch item we could use totalSize to check if we have all possible items or not.~~

### DIFF
--- a/plexapi/base.py
+++ b/plexapi/base.py
@@ -142,19 +142,21 @@ class PlexObject(object):
         clsname = cls.__name__ if cls else 'None'
         raise NotFound('Unable to find elem: cls=%s, attrs=%s' % (clsname, kwargs))
 
-    def fetchItems(self, ekey, cls=None, **kwargs):
+    def fetchItems(self, ekey, cls=None, container_start=None, container_size=None, **kwargs):
         """ Load the specified key to find and build all items with the specified tag
             and attrs. See :func:`~plexapi.base.PlexObject.fetchItem` for more details
             on how this is used.
 
-            Use container_start and container_size for pagination.
+            Parameters:
+                container_start (None, int): offset to get a subset of the data
+                container_size (None, int): How many items in data
+
         """
         url_kw = {}
-        for key, value in dict(kwargs).items():
-            if key == "container_start":
-                url_kw["X-Plex-Container-Start"] = kwargs.pop(key)
-            if key == "container_size":
-                url_kw["X-Plex-Container-Size"] = kwargs.pop(key)
+        if container_start is not None:
+            url_kw["X-Plex-Container-Start"] = container_start
+        if container_size is not None:
+            url_kw["X-Plex-Container-Size"] = container_size
 
         if ekey is None:
             raise BadRequest('ekey was not provided')
--- a/plexapi/library.py
+++ b/plexapi/library.py
@@ -378,7 +378,7 @@ class LibrarySection(PlexObject):
             raise BadRequest('ekey was not provided')
         data = self._server.query(ekey, params=url_kw)
 
-        self._total_size = int(data.attrib.get("totalSize"))
+        self._total_size = utils.cast(int, data.attrib.get("totalSize"))
         items = self.findItems(data, cls, ekey, **kwargs)
 
         librarySectionID = data.attrib.get('librarySectionID')
--- a/plexapi/library.py
+++ b/plexapi/library.py
@@ -381,7 +381,10 @@ class LibrarySection(PlexObject):
         data = self._server.query(ekey, params=url_kw)
 
         if '/all' in ekey:
-            self._total_size = utils.cast(int, data.attrib.get("totalSize"))
+            # totalSize is only included in the xml response
+            # if container size is used.
+            total_size = data.attrib.get("totalSize") or data.attrib.get("size")
+            self._total_size = utils.cast(int, total_size)
 
         items = self.findItems(data, cls, ekey, **kwargs)
 
@@ -585,6 +588,7 @@ class LibrarySection(PlexObject):
 
         results = []
         subresults = []
+        offset = container_start
 
         if maxresults is not None:
             container_size = min(container_size, maxresults)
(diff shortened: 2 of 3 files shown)


Hellowlol:
@Montellese Does solve your issue with search?

plexapi/library.py
```diff
@@ -505,7 +534,8 @@ def listChoices(self, category, libtype=None, **kwargs):
         key = '/library/sections/%s/%s%s' % (self.key, category, utils.joinArgs(args))
         return self.fetchItems(key, cls=FilterChoice)
 
-    def search(self, title=None, sort=None, maxresults=999999, libtype=None, **kwargs):
+    def search(self, title=None, sort=None, maxresults=999999,
+               libtype=None, container_size=X_PLEX_CONTAINER_SIZE, **kwargs):
         """ Search the library. If there are many results, they will be fetched from the server
             in batches of X_PLEX_CONTAINER_SIZE amounts. If you're only looking for the first <num>
```
Montellese, line 540: Maybe adjust the documentation here to mention `container_size`

plexapi/library.py
```diff
@@ -549,16 +580,25 @@ def search(self, title=None, sort=None, maxresults=999999, libtype=None, **kwarg
...
             args['type'] = utils.searchType(libtype)
-        # iterate over the results
-        results, subresults = [], '_init'
+
+        results = []
+        subresults = []
         args['X-Plex-Container-Start'] = 0
-        args['X-Plex-Container-Size'] = min(X_PLEX_CONTAINER_SIZE, maxresults)
-        while subresults and maxresults > len(results):
+        args['X-Plex-Container-Size'] = container_size
```
Montellese, line 587: Doesn't `maxresults` still apply? If `maxresults` is smaller than `container_size` (or implicitly `X_PLEX_CONTAINER_SIZE`) then you are currently retrieving too many items.

plexapi/library.py
```diff
@@ -549,16 +580,25 @@ def search(self, title=None, sort=None, maxresults=999999, libtype=None, **kwarg
...
+        while True:
             key = '/library/sections/%s/all%s' % (self.key, utils.joinArgs(args))
             subresults = self.fetchItems(key)
-            results += subresults[:maxresults - len(results)]
+            results.extend(subresults)
+
+            # self.totalSize is not used as a condition in the while loop as
+            # this require a additional http request.
+            # self.totalSize is updated from .fetchItems
+            if self.totalSize <= len(results):
```
Montellese, line 596: IMO you should take `maxresults` into account here as well. Consider the use case where a library contains 5000 items and the user wants to retrieve the first 1000 in batches of 50 items. That would require 20 requests. Now you are making 100 requests and then throw away 4000 of the items.

Maybe something like this:
```python
maxresults = min(maxresults, self.totalSize)`
if maxresults <= len(results)
    break

args['X-Plex-Container-Start'] += args['X-Plex-Container-Size']
args['X-Plex-Container-Size'] = min(args['X-Plex-Container-Size'], maxresults - len(results))
```
This makes sure that you retrieve exactly the number of items requested. In case the user wants to retrieve all items you perform a few unnecessary calculations but those shouldn't be expensive compared to the (potentially unnecessary) HTTP requests.

And then you also don't need to slice the `results` list at the end.

plexapi/library.py
Hellowlol: I have never used the maxresults argument, so i dropped the ball on this. I don't like the maxresults default value so I'm gonna change this.

plexapi/library.py
Hellowlol: 👍

plexapi/library.py
Montellese: Yeah I don't like that default value either.

plexapi/library.py
```diff
@@ -549,16 +580,25 @@ def search(self, title=None, sort=None, maxresults=999999, libtype=None, **kwarg
             args['sort'] = self._cleanSearchSort(sort)
         if libtype is not None:
             args['type'] = utils.searchType(libtype)
-        # iterate over the results
-        results, subresults = [], '_init'
+
+        results = []
+        subresults = []
         args['X-Plex-Container-Start'] = 0
```
Montellese, line 586: I just realized: With the introduction of `container_start` in #464 do we get a conflict here? This call specifies `X-Plex-Container-Start` to `0` but will this be overwritten by `PlexObject.fetchItems()`? And if not should we adjust this implementation to consider `container_start` as well to get the full flexibility?

Hellowlol: Nope, container_size and container_start in fetchItems is optional. Lets just reuse the entire thing

Hellowlol:
I think this should be good to go. @Montellese ?

plexapi/base.py
```diff
@@ -142,19 +142,21 @@ def fetchItem(self, ekey, cls=None, **kwargs):
         clsname = cls.__name__ if cls else 'None'
         raise NotFound('Unable to find elem: cls=%s, attrs=%s' % (clsname, kwargs))
 
-    def fetchItems(self, ekey, cls=None, **kwargs):
+    def fetchItems(self, ekey, cls=None, container_start=None, container_size=None, **kwargs):
```
Montellese, line 145: How is it possible now to pass `container_start` and `container_size` through e.g. `LibrarySection.all()`? Before it was possible through `**kwargs` but now it would have to be passed through explicitly which is currently not the case.

plexapi/library.py
```diff
@@ -549,15 +582,32 @@ def search(self, title=None, sort=None, maxresults=999999, libtype=None, **kwarg
...
+            # self.totalSize is not used as a condition in the while loop as
+            # this require a additional http request.
+            # self.totalSize is updated from .fetchItems
+            if maxresults is not None:
+                res = min(maxresults, self.totalSize)
+                container_size = min(container_size, maxresults - len(results))
+            else:
+                res = self.totalSize
+
+            if res <= len(results):
```
Montellese, line 606: I think this condition is missing special handling for a request with an initial `container_start > 0` because then the total result set is smaller than `self.totalSize`. So if your library has 1000 items and you start with `container_start = 100` you only expect to retrieve 900 items.

This might actually result in an endless loop because you increase the `container_start` which will probably result in Plex not returning any items if `container_start > totalSize` so `results` will not grow anymore and this condition will never be valid.

tests/test_library.py
```diff
@@ -260,3 +260,6 @@ def test_crazy_search(plex, movie):
     ), "Unable to search movie by year."
     assert movie not in movies.search(year=2007), "Unable to filter movie by year."
     assert movie in movies.search(actor=movie.actors[0].tag)
+    assert len(movies.search(container_start=2, maxresults=1)) == 1
```
Montellese, line 263: Maybe add a test for the case mentioned above witt
```
assert len(movies.search(container_start=2, container_size=1)) == 2
```

plexapi/base.py
Hellowlol: Kwargs can replace keyword arguments. We still test for this.

plexapi/library.py
Hellowlol: Yeah, let just break the loop if the subresult is empty. 👍

tests/test_library.py
Hellowlol: Yep, at att test for container_size higher then total items, atm this will be a endless loop.

Hellowlol:
@Montellese Wanna take a look before i hit merge?

Issue #435, opened 2020-02-27
playMedia() on Nvidia Shield causes subscription failures


jjlawren:
When sending a `Client().playMedia()` command, it seems that [this command](https://github.com/pkkid/python-plexapi/blob/c3e16b2d369968433048ae9a5678294041eb084d/plexapi/client.py#L479) (which only specifies the _server’s_ port) will cause the Plex app on Nvidia Shield devices to treat the `plexapi` request source as an actively listening controller. The Shield client will send POST requests back to the `plexapi` host on the server’s port which repeatedly fail.

Example logs from Shield:
```
02-26 21:58:39.517  i: [pms] /10.6.9.55:57992 - GET /player/timeline/subscribe
02-26 21:58:39.517  i: [Now Playing] New subscriber 0xcfe9033bf0a4 at 10.6.9.55:32400 with command id 23
02-26 21:58:39.519  i: Fetching [method:POST] http://10.6.9.55:32400/:/timeline?X-Plex-Token=...HBKU&includeExternalMedia=1
02-26 21:58:39.525  i: Time out fetching http://10.6.9.55:32400/:/timeline?X-Plex-Token=...HBKU.
02-26 21:58:39.667  i: [pms] /10.6.9.55:57996 - GET /player/playback/playMedia
02-26 21:58:43.325  i: Fetching [method:POST] http://10.6.9.55:32400/:/timeline?X-Plex-Token=...HBKU&includeExternalMedia=1
02-26 21:58:43.333  i: Time out fetching http://10.6.9.55:32400/:/timeline?X-Plex-Token=...HBKU.
02-26 21:58:44.383  i: Fetching [method:POST] http://10.6.9.55:32400/:/timeline?X-Plex-Token=...HBKU&includeExternalMedia=1
02-26 21:58:44.395  i: Time out fetching http://10.6.9.55:32400/:/timeline?X-Plex-Token=...HBKU.
02-26 21:58:45.382  i: Fetching [method:POST] http://10.6.9.55:32400/:/timeline?X-Plex-Token=...HBKU&includeExternalMedia=1
02-26 21:58:45.386  i: Time out fetching http://10.6.9.55:32400/:/timeline?X-Plex-Token=...HBKU.
02-26 21:58:46.384  i: Fetching [method:POST] http://10.6.9.55:32400/:/timeline?X-Plex-Token=...HBKU&includeExternalMedia=1
```

jjlawren:
[...]

PR #489, opened 2020-05-12
Remove 'timeline/subscribe' calls

jjlawren:
Documented [here](https://github.com/plexinc/plex-media-player/wiki/Remote-control-API#new-remote-control-commands) and briefly described in [this comment](https://github.com/pkkid/python-plexapi/issues/435#issuecomment-592058172), setting up subscriptions using `timeline/subscribe` will always "fail" as it's requesting the client to send callbacks to the requester on the provided port. We never create a listener to handle the async timeline updates so they always get rejected at the socket level.

The original workaround for `ptp` and `Plex Media Player` re-subscription intervals have been left intact.

### DIFF
--- a/plexapi/client.py
+++ b/plexapi/client.py
@@ -199,7 +199,7 @@ class PlexClient(PlexObject):
             self._last_call = t
         elif t - self._last_call >= 80 and self.product in ('ptp', 'Plex Media Player'):
             self._last_call = t
-            self.timeline()
+            self.timeline(wait=0)
 
         params['commandID'] = self._nextCommandId()
         key = '/player/%s%s' % (command, utils.joinArgs(params))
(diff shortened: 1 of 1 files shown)


plexapi/client.py
```diff
@@ -548,7 +540,7 @@ def setStreams(self, audioStreamID=None, subtitleStreamID=None, videoStreamID=No
 
     # -------------------
     # Timeline Commands
-    def timeline(self, wait=1):
```
Hellowlol, line 551: I want the default argument to be the same as before. Leave this 1 and just call it with 0 where you need it,

Hellowlol:
What clients have you tested this against?

jjlawren:
Plex Web, Shield, Plex desktop, iOS app. I haven't noticed any regressions in behavior.

Note that I only detected issues when analyzing the Shield client logs. It doesn't cause any visible functional issues as the clients seem to treat this as a non-fatal error.

Hellowlol:
Fix comment i made and this will be included in 4.0

PR #498, opened 2020-05-23
intro_marker

blacktwin:
Adding support for the new Intro Marker for episode class.
Marker class's id and tag are purely cosmetic.


blacktwin:
Currently `reload()` doesn't really exist for Season or Show but there additional attributes we can pull when doing so. Season and Show both return tags for Preferences and On Deck. I'd like to see about adding these into this PR.

Hellowlol:
Nice. I planned to add this. My plan was reuse some bw_plex code to add this as chapter s

blacktwin:
Hahaha yeah I wanted to beat anyone else to it. Was my take sufficient?

Hellowlol:
I’m not sure. I think so but I havnt looked at the xml yet.

blacktwin:
Applying the `_include` string to a shows key gives use the show preferences, hub items (similar items, more from studio and/or writer/director/etc.) and what is OnDeck for this show. If show is not watched than the first episode is what is OnDeck.

amazingr4b:
Nice work getting this potentially added so fast. I put these changes in my setup, and made use of them in the process listener to automatically skip the intros. Has worked well for me so for.

blacktwin:
```python
for episode in plex.library.section('TV Shows Anime').search(libtype='episode'):
    if episode.hasIntroMarker:
        print(episode.title + ' ready to skip intro')
    else:
        print(episode.title + ' analyzing...')
        episode.analyze()
```
Not sure it'd be a good idea to do that but 🤷

plexapi/video.py
```diff
@@ -711,6 +740,13 @@ def seasonEpisode(self):
         """ Returns the s00e00 string containing the season and episode. """
         return 's%se%s' % (str(self.seasonNumber).zfill(2), str(self.index).zfill(2))
 
+    @property
+    def hasIntroMarker(self):
+        """ Returns True if this episode has an intro marker in the xml. """
+        if not self.isFullObject():
+            self.reload()
+        return bool(self.markers)
```
JonnyWong16, line 748: I would only check for `intro` markers in case Plex decides to add different marker types in the future.

```python
any(marker.type == 'intro' for marker in self.markers)
```

Hellowlol:
Anything else you want to add in this?

plexapi/video.py
```diff
@@ -712,6 +741,13 @@ def seasonEpisode(self):
         """ Returns the s00e00 string containing the season and episode. """
         return 's%se%s' % (str(self.seasonNumber).zfill(2), str(self.index).zfill(2))
 
+    @property
+    def hasIntroMarker(self):
+        """ Returns True if this episode has an intro marker in the xml. """
+        if not self.isFullObject():
+            self.reload()
+        return any(marker.type == 'intro' for marker in self.markers)
```
Hellowlol, line 749: Should this be all(), dunno if plex plans to add credits detection.

Issue #504, opened 2020-06-03
Intro marker question


amazingr4b:
**Describe the issue**
"AttributeError: 'NoneType' object has no attribute 'split'" when getting the intro markers for an episode with intro details. 
Not sure if this is a problem with how I am getting these values, or if this is a bug. 

**Code snipppets**
```
shows = plex.library.section(getshowsection())
theshow = shows.get(show).episodes()
for ep in theshow:
	if ((int(ep.seasonNumber) == int(Ssn)) and (int(ep.index) == int(Epnum))):
                for m in ep.markers:
		    if m.type == "intro":
				start = m.start
				end = m.end
				dur = end - start
				start = start / 1000
				end = end / 1000
				dur = dur / 1000
```
throws error:
File "C:\Users\amazi\Desktop\TBN-Plex v517\ToBePosted\plexapi\media.py", line 718, in _loadData
    _tag, _id = self.filter.split('=')
AttributeError: 'NoneType' object has no attribute 'split'

**Expected behavior**
No attribute error. 

I changed the _loadData function in the media.py file to be:
```
 def _loadData(self, data):
	self._data = data
	self.filter = data.attrib.get('filter')
	self.type = data.attrib.get('type')
	if (self.type != "intro"):
		_tag, _id = self.filter.split('=')
		self.tag = self.type + _tag.capitalize()
		self.id = _id
	self.start = cast(int, data.attrib.get('startTimeOffset'))
	self.end = cast(int, data.attrib.get('endTimeOffset'))
```

I no longer get that attribute error, and I get the into details as expected. Not sure if that change would affect anything else, though for the past week and a half I have not encountered any other issues related to this change.  

**Enviroment (please complete the following information):**
 - OS: [ Linux, Windows]
 - Plex version [1.19.4.2865]
 - Python Version [Python 3.8.0]
 - PlexAPI version [4.0.0]

**Additional context**
Add any other context about the issue here.

Hellowlol:
I dont know why this happens. You should try to use episode(season=Xx, episode=ty). The only think i can think of is that the episode dont have the info it needs. You can check using hasIntroMarker (it will reload the object for you if needed)

amazingr4b:
Before I posted my comment I tried using the hasIntroMarker first, and got the same results. As a second test just now I tried:

`	shows = plex.library.section(getshowsection())
	ep = shows.get(show).episode(season=int(Ssn), episode=int(Epnum))
	print (ep.hasIntroMarker)`
[...]

PR #505, opened 2020-06-05
Update Marker class

blacktwin:
Addressing #504 

remove field attribute as it is no longer available.
adding `__repr__` to Marker class resulting in `<Marker:intro:139770:230481>`

### DIFF
--- a/plexapi/media.py
+++ b/plexapi/media.py
@@ -704,13 +704,15 @@ class Marker(PlexObject):
     """
     TAG = 'Marker'
 
+    def __repr__(self):
+        name = self._clean(self.firstAttr('type'))
+        start = self._clean(self.firstAttr('start'))
+        end = self._clean(self.firstAttr('end'))
+        return '<%s>' % ':'.join([p for p in [self.__class__.__name__, name, start, end] if p])
+
     def _loadData(self, data):
         self._data = data
-        self.filter = data.attrib.get('filter')
         self.type = data.attrib.get('type')
-        _tag, _id = self.filter.split('=')
-        self.tag = self.type + _tag.capitalize()
-        self.id = _id
         self.start = cast(int, data.attrib.get('startTimeOffset'))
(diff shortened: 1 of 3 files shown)


plexapi/media.py
```diff
@@ -704,13 +704,15 @@ class Marker(PlexObject):
     """
     TAG = 'Marker'
 
+    def __repr__(self):
+        name = self._clean(self.firstAttr('type'))
+        start = self._clean(self.firstAttr('start'))
+        end = self._clean(self.firstAttr('end'))
+        return '<%s>' % ':'.join([p for p in [self.__class__.__name__, name, start, end] if p])
```
Hellowlol, line 711: Since we are adding a repr, lets show the start and end in "human" time.

````
def sec_to_hh_mm_ss(seconds):
    return time.strftime("%H:%M:%S", time.gmtime(seconds))
````

Comment (blacktwin) [reply plexapi/media.py:711]: Seconds: `[<Marker:intro:139770:230481>]`
Human:  `[<Marker:intro:14:49:30:16:01:21>]`

Comment (blacktwin) [reply plexapi/media.py:711]: Looks like it's actually milliseconds. I've adjusted the code to fit. 

```python
def mill_sec_to_hh_mm_ss(milliseconds):
    if not isinstance(milliseconds, int):
        milliseconds = cast(int, milliseconds)
    hhmmsssm = timedelta(milliseconds=milliseconds)
    return str(hhmmsssm).split(".")[0]
```

Hellowlol: Cant you change it to this one, make the format more correct and is more precice
````
def ms_to_hh_mm_ms(ms):
    r = datetime.datetime.fromtimestamp(ms/1000)
    f = r.strftime("%H:%M:%S.%f")
    return f[:-2]
````

Hellowlol: Also Change the repr so its [<Marker:intro: 00:02:19 - 00:03:50>]

PR #509, opened 2020-06-05
Add fields to most objects

blacktwin:
Adding fields attributes to objects that would have them.
Fields will help with finding what metatdata fields have been edited and locked.


plexapi/library.py
```diff
@@ -1085,6 +1085,29 @@ def __len__(self):
...
+            ratingKey (int): Unique key identifying this item.
+            addedAt (datetime): Datetime this item was added to the library.
+            childCount (int): Count of child object(s)
+            collectionMode (str): How the items in the collection are displayed.
+            collectionSort (str): How to sort the items in the collection.
+            fields (list): List of :class:`~plexapi.media.Field`.
+            key (str): API URL (/library/metadata/<ratingkey>).
+            index (int): Unknown
+            maxYear (int): Year
+            minYear (int): YEar
```
JonnyWong16, line 1103: "Year"

plexapi/library.py
```diff
@@ -1085,6 +1085,29 @@ def __len__(self):
...
+            subtype (str): Media type
+            summary (str): Summary of the collection
+            thumb (str): URL to thumbnail image.
+            title (str): Artist, Album or Track title. (Jason Mraz, We Sing, Lucky, etc.)
```
JonnyWong16, line 1107: Title of the collection.

plexapi/library.py
```diff
@@ -1093,20 +1116,21 @@ class Collections(PlexObject):
     def _loadData(self, data):
         self.ratingKey = utils.cast(int, data.attrib.get('ratingKey'))
         self._details_key = "/library/metadata/%s%s" % (self.ratingKey, self._include)
+        self.addedAt = utils.toDatetime(data.attrib.get('addedAt'))
+        self.childCount = utils.cast(int, data.attrib.get('childCount'))
+        self.collectionMode = data.attrib.get('collectionMode')
+        self.collectionSort = data.attrib.get('collectionSort')
```
JonnyWong16, line 1122: Maybe for a different PR:

For some collections, I don't see a `collectionMode` or `collectionSort` in my XML, even after changing the collection settings.

plexapi/library.py
```diff
@@ -1093,20 +1116,21 @@ class Collections(PlexObject):
...
         self.thumb = data.attrib.get('thumb')
-        self.addedAt = utils.toDatetime(data.attrib.get('addedAt'))
+        self.title = data.attrib.get('title')
+        self.type = data.attrib.get('type')
         self.updatedAt = utils.toDatetime(data.attrib.get('updatedAt'))
-        self.childCount = utils.cast(int, data.attrib.get('childCount'))
-        self.minYear = utils.cast(int, data.attrib.get('minYear'))
-        self.maxYear = utils.cast(int, data.attrib.get('maxYear'))
-        self.collectionMode = data.attrib.get('collectionMode')
-        self.collectionSort = data.attrib.get('collectionSort')
```
JonnyWong16, line 1134: Maybe for a different PR:

Add attributes `guid`, `titleSort`, `librarySectionTitle`, `librarySectionID`, `contentRating`, `summary`.

plexapi/library.py
blacktwin: Looks like `includePreferences=1` is needed in your url.

plexapi/library.py
blacktwin: added

blacktwin:
Needs tests. Edit something, lock it, reload, assert field is locked.

plexapi/library.py
```diff
@@ -1084,7 +1084,38 @@ def __len__(self):
 
 
 @utils.registerPlexObject
-class Collections(PlexObject):
+class Collections(PlexPartialObject):
```
Hellowlol, line 1087: Was this changed to PlexPartialObject to allow a reload?

blacktwin: 9685b57   
> this will allow collections to use the base.PlexPartialObject.edit method

More so that I can use the `.edit()`

Issue #462, opened 2020-04-22
Add an actor to a movie/video with role and thumb


dominikpflegerl:
So far I have created a little script to add an specific actor to a search result:

```
from plexapi.server import PlexServer
PLEX_URL = 'http://192.168.178.100:32400'
PLEX_TOKEN = 'token'
homeserver = PlexServer(PLEX_URL, PLEX_TOKEN)

tag = input(f"Actor: ")

for video in homeserver.library.section('test').search(None, collection=[tag]): 
	video._edit_tags(tag="actor", items=[tag])
```
It work's fine. An actor appears on that particular video(s) in the collection.
Now, I just don't want to add the actor, I also want to add the role and a thumb (photo URL), but so far I could not find anything in the documentation and in the source code. (might be it's quite late and I am tired), but I would be thankful if someone could help me with it or make it possible.

blacktwin:
You and me both. Creating an actor also creates the actor in the DB where I assume you could add the missing thumb and role but I haven't tried that yet. I'd prefer to do it all with API commands.

blacktwin:
@JonnyWong16 found how to add a thumb.

```python
edits = {
    'actor[0].tag.tag': 'Mom',
    'actor[0].locked': 1,
    'actor[0].tag.thumb': 'https://nickelodeonuniverse.com/wp-content/uploads/Spongebob.png'
}
movie = plex.library.section('Movies').get('Title')
movie.edit(**edits)
```
 Still trying to find role.

JonnyWong16:
Found it.

```
edits = {
    'actor[0].tag.tag': 'Mom',
    'actor[0].locked': 1,
    'actor[0].tagging.text': 'Test Role',
    'actor[0].tag.thumb': 'https://nickelodeonuniverse.com/wp-content/uploads/Spongebob.png'
}
movie = plex.library.section('Movies').get('Title')
movie.edit(**edits)
```
![image](https://user-images.githubusercontent.com/9099342/80121264-f00ce700-8540-11ea-83f5-5687cb640de9.png)

blacktwin:
Need a addActor or add add additional kwargs if it already exists.

Hellowlol:
Imo we should support objects. So Role.create() and allow that to be passed to edit that serialize this to a url.

blacktwin:
Agreed.

JonnyWong16:
[...]

PR #510, opened 2020-06-06
Actor editing

blacktwin:
Addressing Issue #462 

addActor() and removeActor() methods are now added to video.Movie and video.Show.

### DIFF
--- a/plexapi/video.py
+++ b/plexapi/video.py
@@ -359,6 +359,15 @@ class Movie(Playable, Video):
         return filepaths
 
     def addActor(self, index=0, name=None, role=None, thumb=None, locked=True):
+        """ Add an Actor.
+
+           Parameters:
+                index (int): Positional index
+                name (str): Name of Actor.
+                role (str): Role actor plays in video.
+                thumb (str): URL to image file.
+                locked (bool): True = 1, False = 0
+        """
         edits = {}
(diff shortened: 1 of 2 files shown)


plexapi/video.py
```diff
@@ -358,6 +358,37 @@ def download(self, savepath=None, keep_original_name=False, **kwargs):
...
+        if role:
+            edits['%s.tagging.text' % actor] = role
+        if thumb:
+            edits['%s.tag.thumb' % actor] = thumb
+        if locked:
+            edits['%s.locked' % actor] = int(locked)
+        self.edit(**edits)
+
+    def removeActor(self, name):
+        """ Add an Actor from item. """
```
JonnyWong16, line 384: "Remove"

plexapi/video.py
```diff
@@ -528,6 +559,36 @@ def download(self, savepath=None, keep_original_name=False, **kwargs):
...
```
JonnyWong16, line 585: "Remove"

blacktwin:
> Question: Does adding an actor with the same name that already exists in a different movie use the same role `id` and `filter`?

Yes.

blacktwin:
`addActor()` can also be used to edit any existing actor's thumb. The `name` is the only required argument to make changes. `name` and whatever other attribute someone would want to change.

PR #512, opened 2020-06-09
Video optimize update

blacktwin:
Initial PR for video.optimize had an incorrect assumption for the locationID usage. locationID uses the library's path id to determine where to place the optimized version. 

Added a check inside the method to only allow library locations based on the library location ID. Library location ID was initially missing from library.locations as this attribute only pulled the path and not the ID. Now location is it's own class and includes both ID and path.


Hellowlol:
This would be a breaking change as we change frome a str to a class for the location.

blacktwin:
Could repeat the listAttrib for a locationsID attribute.

PR #518, opened 2020-06-22
Library Hubs and Music Stations

blacktwin:
Adding Station support as requested in #361

Adding Library Hubs. Returns "More By {ARTIST}" or "Rediscover"

- [x] New feature (non-breaking change which adds functionality)

PR #519, opened 2020-06-28
library:  add unmatched as a filter for Movies and Show Sections

mbjurstrom:
This add unmatched as a filter for the Movies and Show sections. 
I would expect it to also work for the Music and Photo section, but I haven't added that as I don't have any photos or music in my Plex libraries.


blacktwin:
This maybe obsolete if PR #518 is merged as commit ea6538f would remove the hardcoding of available filters and commit a665d6f adds a dynamic way to populate the available filters. Just a heads up.

mbjurstrom:
That is actually a cool update. Thanks for the heads up. 
If that ever get merged. I guess they can close this PR. 
I only added the unmatched filter as I needed it and found it to be missing.

blacktwin:
#518 was merged which makes this PR obsolete. Closing.

PR #526, opened 2020-07-12
Fixes #525: Added package metadata for supported Python versions

### LINKED ISSUE #525: PlexAPI 4.0.0 removed support for py27 but still installs successfully on py27
**Describe the issue**

Installing PlexAPI 4.0.0 on Python 2.7 succceeds:

```
> python --version
Python 2.7.15
> pip install plexapi==4.0.0
> pip list | grep -i plexapi
PlexAPI            4.0.0
```

However, using it then fails:
```
> python -c "import plexapi"
  File "<string>", line 1, in <module>
  File "C:\Users\Andi\Envs\plexmediafixup\lib\site-packages\plexapi\__init__.py", line 9, in <module>
    from plexapi.utils import SecretsFilter
  File "C:\Users\Andi\Envs\plexmediafixup\lib\site-packages\plexapi\utils.py", line 10, in <module>
    from urllib.parse import quote
ImportError: No module named parse
```

`urllib.parse` is a new module in Python 3, where the functions of `urllib` were split into submodules.
The code that fails was added in PlexAPI 4.0.0. PlexAPI versions 3.6.0 and below run fine on Python 2.7.

So essentially, PlexAPI 4.0.0 removed support for Python 2.7, but it does not indicate that anywhere in its package on Pypi.

I am fully aware that Python 2.7 has reached its EOL. The issue is that PlexAPi 4.0.0 successfully installs on Python 2.7, and then fails.

**Code snipppets**

See above

**Expected behavior**

I am expecting that PlexAPI indicates in its package on Pypi which Python versions it supports, so that there is no discrepancy between successful installation and successful use.

**Enviroment (please complete the following information):**
 - OS: Windows
 - Plex version: latest
 - Python Version: 2.7.16
 - PlexAPI version: 4.0.0

**Additional context**

None

andy-maier:
See commit message for details.
Ready for review and merge.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -4,7 +4,10 @@
 Install PlexAPI
 """
 import re
-from setuptools import setup
+try:
+    from setuptools import setup
+except ImportError:
+    from distutils.core import setup
 
 # Get the current version
 with open('plexapi/__init__.py') as handle:
--- a/setup.py
+++ b/setup.py
@@ -4,7 +4,7 @@
 Install PlexAPI
 """
 import re
-from distutils.core import setup
+from setuptools import setup
 
 # Get the current version
 with open('plexapi/__init__.py') as handle:
@@ -32,6 +32,11 @@ setup(
     url='https://github.com/pkkid/python-plexapi',
     packages=['plexapi'],
     install_requires=requirements,
+    python_requires='>=3.5',
     long_description=readme,
     keywords=['plex', 'api'],
+    classifiers=[
+        'Operating System :: OS Independent',
+        'Programming Language :: Python :: 3',
+    ]
 )


setup.py
```diff
@@ -4,7 +4,7 @@
 Install PlexAPI
 """
 import re
-from distutils.core import setup
+from setuptools import setup
```
Hellowlol, line 7: Revert this or try exception import error with distutils as fallback. Setuptools isnt part of the stdlib.

andy-maier: The python_requires argument needed by Pypi is only supported with setuptools, but not with distutils. So if we fall back on distutils, the keyword could not be specified. Which could of course be done.

However, consider that the Python packaging authority recommends using setuptools over distutils for several years now. See here: https://packaging.python.org/guides/tool-recommendations/#packaging-tool-recommendations

I have not yet seen any pre-installed Python environment on any OS (e.g. Windows, CentOS, Ubuntu, macOS) that does not have setuptools out of the box for years now. All tools creating virtual Python environments install setuptools automatically, and most packages on Pypi depend on setuptools without specifying it in their package requirements.

Let me know whether you still want me to add a fallback to distutils.

PR #541, opened 2020-08-03
Improve clips handling

jjlawren:
Some minor changes to populate the available attributes on `clip` media types, mostly by using the parent class methods.

Used an override method for `section()` as `librarySectionID` doesn't seem to be populated in any of the `clip` session examples I reviewed. Will likely always return `None`. If it _does_ show up for someone else or gets added in the future it should return the proper library section.

### DIFF
--- a/plexapi/video.py
+++ b/plexapi/video.py
@@ -787,10 +787,9 @@ class Clip(Playable, Video):
         self.duration = utils.cast(int, data.attrib.get('duration'))
         self.extraType = utils.cast(int, data.attrib.get('extraType'))
         self.guid = data.attrib.get('guid')
+        self.index = utils.cast(int, data.attrib.get('index'))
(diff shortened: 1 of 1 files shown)


jjlawren:
This works, but I do notice that the `librarySectionID` attribute is forcibly reloaded somewhat often:
```
Reloading Clip 'Final_Trailer' for attr 'librarySectionID'
```
Is there a clean way to mark this attribute for this specific class as non-reloadable?
https://github.com/pkkid/python-plexapi/blob/82df456f7a38aa17e8c63e0b342c992d3a889d30/plexapi/base.py#L294-L309

jjlawren:
It's possible to avoid reloading by creating a `_librarySectionId` attribute or similar. However, I think that would be unnecessarily messy when this key doesn't seem exist in any example I've seen so far. Let's just ignore it completely for now.

Issue #560, opened 2020-08-29
Missing 'continuous' arg on PlayQueue.create


kylezimmerman:
## Description
The `PlayQueue.create()` function has several args already (`shuffle=0, repeat=0, includeChapters=1, includeRelated=1`), but is missing the `continuous` arg.

## Why
When `continuous=1` is set, the `PlayQueue` will be created starting at at the provided `item`, but will include the next episodes in a show if item is a show. It appears to do nothing if the item is a movie. When `continuous=0` or it is omitted, it just adds the single episode and then stops playing (current behavior).

This would let you easily queue up a show starting at a particular episode and continuing from there.

kylezimmerman:
Submitted PR https://github.com/pkkid/python-plexapi/pull/561 to fix this

PR #561, opened 2020-08-29
PlayQueue: Add optional continuous arg to PlayQueue.create()

kylezimmerman:
## Description
Add an optional `continuous` arg to the `PlayQueue.create()` function. It defaults to `continuous=0` so that the current behaviour remains the default so it's non-breaking.

## Why
This allows to much more easily queue up a an entire show starting at a particular episode and continuing from there, which is the behavior on most Plex clients.

- When `continuous=1` is set, the `PlayQueue` will be created starting at at the provided `item`, but will also automatically include the next episodes in a show if item is a show. It appears to do nothing if the item is a movie.
- When `continuous=0` or it is omitted, it just adds the single episode and then stops playing (current behavior).

## Examples

Implicit (continuous=0 by default):
```py
play_queue = PlayQueue.create(plex_server, example_tv_show_s03e02)
# play_queue.items has one item [example_tv_show_s03e02]
```

Explicit `continuous=0`:
```py
play_queue = PlayQueue.create(plex_server, example_tv_show_s03e02, continuous=0)
# play_queue.items has one item [example_tv_show_s03e02]
```

Explicit `continuous=1`:
```py
play_queue = PlayQueue.create(plex_server, example_tv_show_s03e02, continuous=1)
# play_queue.items has many items, [example_tv_show_s03e02, example_tv_show_s03e03, example_tv_show_s03e04, ...]
```

Fixes https://github.com/pkkid/python-plexapi/issues/560

### DIFF
--- a/plexapi/playqueue.py
+++ b/plexapi/playqueue.py
@@ -42,7 +42,7 @@ class PlayQueue(PlexObject):
         self.items = self.findItems(data)
 
     @classmethod
-    def create(cls, server, item, shuffle=0, repeat=0, includeChapters=1, includeRelated=1):
+    def create(cls, server, item, shuffle=0, repeat=0, includeChapters=1, includeRelated=1, continuous=0):
         """ Create and returns a new :class:`~plexapi.playqueue.PlayQueue`.
 
             Paramaters:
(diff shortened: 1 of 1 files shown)


jjlawren:
Thanks, this works well. I'll rebase this into #563 and fix the flake failure.

PR #562, opened 2020-08-31
 guids collection to movie and GuidTag PlexObject

andyloree:
Suggested implementation for Movie guids collection and GuidTag PlexObject in reference to new Plex movie agent, which is still in beta.

MediaTag's id expects type int, so I figured a specific version might be better.  Might be worth implementing parsing of the expected id's (tmdb or imdb), but for now this fits to the existing tag collection model

### DIFF
--- a/plexapi/media.py
+++ b/plexapi/media.py
@@ -510,6 +510,29 @@ class MediaTag(PlexObject):
             raise BadRequest('Key is not defined for this tag: %s' % self.tag)
         return self.fetchItems(self.key)
 
+class GuidTag(PlexObject):
+    """ Base class for guid tags used only for Guids, as they contain only a string identifier
+
+        Attributes:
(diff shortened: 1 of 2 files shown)


xfouloux:
Hi,

you still can't search by guids, but you can print a Item.guids
any idea how to do the search part ?

Regards.

plexapi/media.py
```diff
@@ -589,6 +612,10 @@ class Genre(MediaTag):
     TAG = 'Genre'
     FILTER = 'genre'
 
+@utils.registerPlexObject
```
blacktwin, line 615: Needs an additional line spacing around this class.

blacktwin:
The test environment would need at least one Movie Library to have the new Plex Movie agent enabled in order to add any testing related to this PR.

glensc:
looking forward to this to get merged. it's a dependency of such or: https://github.com/Taxel/PlexTraktSync/pull/86

PR #563, opened 2020-09-05
Playqueue improvements

jjlawren:
Allows creating playqueues on the Plex server with an initial list of arbitrary items. Also allows appending media to an existing PlayQueue into the "Up Next" area, rearranging items, removing items, and clearing the PlayQueue completely.

Updated documentation, cast attributes to appropriate types, use f-strings where possible, format with `black` for consistency.

### DIFF
--- a/plexapi/playqueue.py
+++ b/plexapi/playqueue.py
@@ -97,6 +97,7 @@ class PlayQueue(PlexObject):
         cls,
         server,
         items,
+        startItem=None,
         shuffle=0,
         repeat=0,
         includeChapters=1,
@@ -109,6 +110,8 @@ class PlayQueue(PlexObject):
             server (:class:`~plexapi.server.PlexServer`): Server you are connected to.
             items (:class:`~plexapi.media.Media` or :class:`~plexapi.playlist.Playlist`):
                 A media item, list of media items, or Playlist.
+            startItem (:class:`~plexapi.media.Media`, optional):
+                Media item in the PlayQueue where playback should begin.
             shuffle (int, optional): Start the playqueue shuffled.
             repeat (int, optional): Start the playqueue shuffled.
             includeChapters (int, optional): include Chapters.
@@ -136,6 +139,9 @@ class PlayQueue(PlexObject):
             uuid = items.section().uuid
             args["type"] = items.listType
             args["uri"] = f"library://{uuid}/item/{items.key}"
+
+        if startItem:
+            args["key"] = startItem.key
 
         path = f"/playQueues{utils.joinArgs(args)}"
         data = server.query(path, method=server._session.post)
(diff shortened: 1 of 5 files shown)


plexapi/playqueue.py
```diff
@@ -58,19 +58,37 @@ def _loadData(self, data):
...
+
     def __contains__(self, media):
         """Returns True if the PlayQueue contains the provided media item."""
-        return any(x == media for x in self.items)
+        return any(x.playQueueItemID == media.playQueueItemID for x in self.items)
 
-    def has_queueItemID(self, playQueueItemID):
-        """Returns True if the PlayQueue contains an item with the provided playQueueItemID."""
-        return any(x.playQueueItemID == playQueueItemID for x in self.items)
+    def get_queue_item(self, item):
```
blacktwin, line 72: Shouldn't this follow the camel casing? `getQueueItem`

jjlawren: Yeah, I'm torn on this. It's not returning a queueItem object, but a regular media item that has been enriched with a `playQueueItemID` attribute for being part of a `PlayQueue`: https://github.com/pkkid/python-plexapi/blob/5b47a3be0c96749e58adbaa4ac7f57c1e14f1c2e/plexapi/base.py#L601

I'm open to whatever feels more logical or consistent.

plexapi/playqueue.py
```diff
@@ -1,77 +1,232 @@
...
+        self.playQueueVersion = utils.cast(int, data.attrib.get("playQueueVersion"))
+        self.size = utils.cast(int, data.attrib.get("size", 0))
         self.items = self.findItems(data)
 
+    def __len__(self):
+        return len(self.items)
+
+    def __iter__(self):
+        for item in self.items:
+            yield item
```
maykar, line 70:
```
    def __iter__(self):
        yield from self.items
```

Is a bit cleaner

plexapi/playqueue.py
```diff
@@ -1,77 +1,232 @@
...
-        args['shuffle'] = shuffle
-        args['continuous'] = continuous
-        if item.type == 'playlist':
-            args['playlistID'] = item.ratingKey
-            args['type'] = item.playlistType
+        args["includeChapters"] = includeChapters
+        args["includeRelated"] = includeRelated
+        args["repeat"] = repeat
+        args["shuffle"] = shuffle
+        args["continuous"] = continuous
```
maykar, line 120:
```
        args = {
            "includeChapters": includeChapters,
            "includeRelated": includeRelated,
            "repeat": repeat,
            "shuffle": shuffle,
            "continuous": continuous,
        }
```
Cleaner as well

plexapi/playqueue.py
maykar, line 70:
```
    def __iter__(self):
        yield from self.items
```
Would be a bit cleaner

plexapi/playqueue.py
jjlawren: Went with your suggestion and made it camel case for consistency.

jjlawren:
I think this is ready. Added tests for 100% coverage on everything added.

maykar (approved):
Put it through its paces, quite a nice addition. Looks good, works great!

Issue #454, opened 2020-04-16
State of plexapi


Hellowlol:
Hello everybody.

plexapi haven't gotten the attention it need for quite some time. PRS have been unmerged, test suite is broken.

Going forward I would like to fix this. I would like to get some input from any stakeholders regarding this. @pkkid @blacktwin @andrey-yantsen @jjlawren

Short time plan:
- [ ] Fix any broken tests
- [ ] Release 3.4.0

Some time in a not so distant future:
- [ ] Review all tests, some are way to flaky
- [ ] Test on newer python version
- [ ] Use black as linter
- [ ] Drop python 2 in 4.0.0 and merge any breaking PRS (#370)

andrey-yantsen:
Good to see that you're still interested in the project, awesome! The plan sounds great as well.

As for tests — one of the issues here is that we're downloading required content for the library for every run, and sometimes it's just failing. Some other times — we're unable to properly prepare the library and understand that it's basically ready...
In my other plex-related project ([plex-api.rs](https://github.com/andrey-yantsen/plex-api.rs/), the API written in Rust) I've gone in slightly another way: the Plex library (the database and all the required media) is stored inside the repo, hence bootstrap is much faster and stable, but I had to sacrifice some disk space to it.

It's just a suggestion, feel free to choose the better path :)

Hellowlol:
@andrey-yantsen thanks for the feedback. Ill stub the files to skip the download. That’s a great idea. I wonder why picked so large files..

andrey-yantsen:
I was young, sorry :D Feel free to steal the database from plex-api.rs, there is no copyright, the media — are just files with white noise.

pkkid:
I like the plan.

To me, getting the tests more reliable is a top priority. If those are running consistently and reliably, we can be more confident and move faster on all other changes. -- I did try to take a stab at fixing some of the flakyness in the tests. It seems during test setup when it downloads and configures the temp Plex client, it doesn't always see or register all media we are giving it. I ended up scraping work in the area because it didn't end up helping anything. I like the idea of storing test media in the repo, it's one less variable we have to worry about not working. -- Testing the tests is an annoying area because we have to push the changes to get Travis to execute the scripts. Maybe it'll feel cleaner if we do this on another branch. 

Not sure what the Black linter is so I have no comments. I always use flake8.

[...]

PR #569, opened 2020-09-14
Add 'activities' endpoint support

jjlawren:
PMS reports back information on its currently running activities, such as scanning for metadata.

This is an MVP version which should allow querying to see if metadata activities are still in progress as suggested [here](https://github.com/pkkid/python-plexapi/issues/454#issuecomment-691329452).

### DIFF
--- a/plexapi/server.py
+++ b/plexapi/server.py
@@ -190,6 +190,12 @@ class PlexServer(PlexObject):
         data = self.query(Account.key)
         return Account(self, data)
 
+    @property
+    def activities(self):
+        """ Returns a list of all server settings. """
+        for elem in self.query(Activities.key):
+            yield Activities(self, elem)
+
     def agents(self, mediaType=None):
         """ Returns the `:class:`~plexapi.media.Agent` objects this server has available. """
         key = '/system/agents'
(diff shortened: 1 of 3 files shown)


plexapi/server.py
```diff
@@ -190,6 +190,14 @@ def account(self):
         data = self.query(Account.key)
         return Account(self, data)
 
+    @property
+    def activities(self):
+        """Returns all current PMS activities."""
+        activities = []
+        for elem in self.query(Activity.key):
+            activities.append(Activity(self, elem))
+        return activities
```
maykar, line 199:
```py
    @property
    def activities(self):
        """Returns all current PMS activities."""
        return [Activity(self, elem) for elem in self.query(Activity.key)]
```

jjlawren:
I've added a test that's intended to avoid some test failures based on incomplete metadata. The tests that rely on metadata agents will now wait to run until all server activities have finished.

I ran a "canary" test in this PR which showed that agents are sometimes still scanning when we begin validating metadata correctness in the CI test runs.

jjlawren:
Moved it to the start of the run (alphabetically) and added an authenticated version so it's guaranteed to run first in both auth and anon runs.

I'm still not sure how this is failing since there's an explicit check in [`plex-bootstraptest.py`](https://github.com/pkkid/python-plexapi/blob/fb82bc402bd725220687b76c01a35008884b75ac/tools/plex-bootstraptest.py#L279-L280) that tries to wait for metadata scans. In the canary test that failed the bootstrap script seemed to believe the scanning was complete. Maybe we need multiple different methods to be sure.

jjlawren:
Well, that theory didn't pan out. After multiple successful runs, the latest CI run passed the activity check but still failed in the audio metadata:
https://travis-ci.org/github/pkkid/python-plexapi/jobs/727760606#L839-L873

jjlawren:
This didn't prevent the metadata test failures, but I don't see a downside to including these tests as a failsafe. Will merge alongside #571 and see how the CI behaves before making more changes to the tests.

PR #572, opened 2020-09-16
Update timeline to return an active timeline object w/ attributes

maykar:
* Updates `timeline` to create and return an active timeline object with associated attributes and caching
* Refactor `isPlayingMedia()` to utilize changes and fix `includePaused` default value (was False when docs stated True)


plexapi/client.py
```diff
@@ -199,7 +201,7 @@ def sendCommand(self, command, proxy=None, **params):
             self._last_call = t
         elif t - self._last_call >= 80 and self.product in ('ptp', 'Plex Media Player'):
             self._last_call = t
-            self.timeline(wait=0)
+            self.timelines()
```
Hellowlol, line 204: I'm not a fan of this change. This will slow down any commands. We should use a raw self.sendCommand('timeline/poll', wait=0) as we don't care about the response. If a user really want the timeline info they should self.timlines() or self.timeline

plexapi/client.py
```diff
@@ -540,20 +542,63 @@ def setStreams(self, audioStreamID=None, subtitleStreamID=None, videoStreamID=No
 
     # -------------------
     # Timeline Commands
-    def timeline(self, wait=1):
-        """ Poll the current timeline and return the XML response. """
-        return self.sendCommand('timeline/poll', wait=wait)
+    def timelines(self, wait=0):
+        """Poll the client's timelines, create, and return timeline objects."""
+        self.sendCommand(ClientTimeline.key, wait=1)
```
maykar, line 547: Rather than using this workaround to "wakeup" client's not reporting timelines, I should just document Plex's buggy behavior for now. Still attempting to find a better way though.

Hellowlol: Imo the default argument should be 1 to increase the chance that we get a proper response timeline response. We don’t need it for work around as the point for the work around is just to make sure the command that we actually want to send is executed by pmp etc. Nice work btw. Thanks.

maykar: Yeah, I think I'm happy with that for now. Will make the change. Thank you!

jjlawren: Having the default be `wait=1` makes a timeline call potentially blocking. I don't think that's a safe default anymore.

Hellowlol: Just do what ever you prefer then. There are some info about the remote control api here: https://github.com/plexinc/plex-media-player/wiki/Remote-control-API (incase you havnt seen it)

maykar: After further testing with this, the workaround and even adding a wait don't solve this issue. Initially it seemed those things helped, but it was a red herring. Was able to repeat the issue even with a wait and multiple workarounds.

It seems to be a bug on Plex's side and rather than add a workaround that causes timeline calls to block (sometimes for quite a bit), it's best to leave it as it was and just document the behavior for now. Will make the changes and document the issue.

plexapi/client.py
```diff
@@ -540,20 +542,65 @@ def setStreams(self, audioStreamID=None, subtitleStreamID=None, videoStreamID=No
...
+        state = getattr(self.timeline, "state", None)
+        return bool(state == 'playing' or (includePaused and state == 'paused'))
+
+
+class ClientTimeline(PlexObject):
+    """Get the timeline's attributes."""
+
+    key = 'timeline/poll'
+
+    def _loadData(self, data):
```
jjlawren, line 579: Some additional attributes I've seen when music is playing:
```
audioStreamID (int)
partCount (int)
partIndex (int)
```

plexapi/client.py
```diff
@@ -540,20 +542,65 @@ def setStreams(self, audioStreamID=None, subtitleStreamID=None, videoStreamID=No
...
+        self._data = data
+        self.address = data.attrib.get('address')
+        self.autoPlay = utils.cast(int, data.attrib.get('autoPlay'))
```
jjlawren, line 582: This is delivered as an `int` but it represents a `bool`. Should we change the casting?

plexapi/client.py
```diff
@@ -540,20 +542,65 @@ def setStreams(self, audioStreamID=None, subtitleStreamID=None, videoStreamID=No
...
+        self.containerKey = data.attrib.get('containerKey')
+        self.controllable = data.attrib.get('controllable')
+        self.duration = utils.cast(int, data.attrib.get('duration'))
+        self.itemType = data.attrib.get('itemType')
+        self.key = data.attrib.get('key')
+        self.location = data.attrib.get('location')
+        self.machineIdentifier = data.attrib.get('machineIdentifier')
+        self.playQueueID = data.attrib.get('playQueueID')
+        self.playQueueItemID = data.attrib.get('playQueueItemID')
```
jjlawren, line 591:
```suggestion
        self.playQueueID = utils.cast(int, data.attrib.get('playQueueID'))
        self.playQueueItemID = utils.cast(int, data.attrib.get('playQueueItemID'))
```

plexapi/client.py
```diff
@@ -540,20 +542,65 @@ def setStreams(self, audioStreamID=None, subtitleStreamID=None, videoStreamID=No
...
+        self.playQueueVersion = utils.cast(int, data.attrib.get('playQueueVersion'))
+        self.port = utils.cast(int, data.attrib.get('port'))
+        self.protocol = data.attrib.get('protocol')
+        self.providerIdentifier = data.attrib.get('providerIdentifier')
+        self.ratingKey = utils.cast(int, data.attrib.get('ratingKey'))
+        self.repeat = utils.cast(int, data.attrib.get('repeat'))
```
jjlawren, line 597: `bool`?

plexapi/client.py
```diff
@@ -540,20 +542,65 @@ def setStreams(self, audioStreamID=None, subtitleStreamID=None, videoStreamID=No
...
+        self.seekRange = data.attrib.get('seekRange')
+        self.shuffle = utils.cast(int, data.attrib.get('shuffle'))
```
jjlawren, line 599: `bool`?

jjlawren (changes requested):

jjlawren (approved):
Thanks, looking good!

PR #573, opened 2020-09-21
Add Library timeline support

jjlawren:
Adds support for querying the `/library/sections/<SECTION_ID>/timeline` endpoint. The useful information here is `updateQueueSize` which represents the number of items in the library section which are still being processed.

The `/activities` calls added in #569 will report the agents/scanners in progress, but it appears that item metadata may be processed asynchronously after those activities have completed. This adds a test which runs before all other tests which rely on stable metadata and ensures these item processing queues are empty before proceeding.

### DIFF
--- a/plexapi/library.py
+++ b/plexapi/library.py
@@ -466,6 +466,12 @@ class LibrarySection(PlexObject):
         data = self._server.query(key)
         return self.findItems(data, cls=Setting)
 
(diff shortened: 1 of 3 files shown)


jjlawren:
I attempted to force failures without luck. If these are going to prevent slow metadata processing it's likely dependent on slow external services.

Issue #591, opened 2020-10-23
Data disappears when interacting with object in iPython


fz6:
**Describe the issue**
This sounds a bit crazy, and honestly I'm very puzzled as to what is going on. When interacting with an `Episode` object in `ipython`, attributes that previously returned data seem to become empty?

**Code snipppets**
```
In [4]: from plexapi.server import PlexServer                                                                                                                                                                                                 

In [5]: plex = PlexServer(baseurl, token)

In [7]: plex.sessions()                                                                                                                                                                                                                       
Out[7]: [<Episode:redacted>]

In [8]: session = plex.sessions()[0]                                                                                                                                                                                                          

# Data is missing already, even though I can see it at API endpoint:

In [9]: session.transcodeSessions                                                                                                                                                                                                             
Out[9]: []

In [10]: session.players                                                                                                                                                                                                                      
Out[10]: []

In [13]: plex.sessions()[0].players                                                                                                                                                                                                           
Out[13]: [<PlexClient:Chrome>]

In [14]: session = plex.sessions()[0]                                                                                                                                                                                                         

In [15]: session.players                                                                                                                                                                                                                      
Out[15]: []

In [16]: session.transcodeSessions                                                                                                                                                                                                            
Out[16]: []

In [17]: plex.sessions()[0].transcodeSessions                                                                                                                                                                                                 
Out[17]: [<TranscodeSession:/transcode/sessions/>]

In [46]: session = plex.sessions()[0]                                                                                                                                                                                                         

In [47]: session.players                                                                                                                                                                                                                      
Out[47]: [<PlexClient:Chrome>]

In [48]: session.sessionKey                                                                                                                                                                                                                   
Out[48]: 3

In [49]: session.players                                                                                                                                                                                                                      
Out[49]: [<PlexClient:Chrome>]

In [50]: session.players                                                                                                                                                                                                                      
Out[50]: [<PlexClient:Chrome>]

In [51]: time.sleep(10)                                                                                                                                                                                                                       

In [52]: session.players                                                                                                                                                                                                                      
Out[52]: [<PlexClient:Chrome>]

In [53]: session.transcodeSessions                                                                                                                                                                                                            
Out[53]: [<TranscodeSession:/transcode/sessions/>]

In [54]: session.session                                                                                                                                                                                                                      
Out[54]: [<Session:redacted>]

In [55]: session.session                                                                                                                                                                                                                      
Out[55]: [<Session:redacted>]

[...]

PR #594, opened 2020-10-27
Don't overwrite existing attributes with an empty list

jjlawren:
As reported in #591, reloading objects to update missing attributes can cause list attributes to be lost. This is easiest to reproduce when accessing a missing attribute on a `Playable` object that was built from a `PlexServer.session()` call. Any existing `players`, `usernames`, `session`, `transcodeSessions`, etc are lost when reloading the object from the associated media item.

### DIFF
--- a/plexapi/base.py
+++ b/plexapi/base.py
@@ -283,7 +283,7 @@ class PlexPartialObject(PlexObject):
     """
 
     def __eq__(self, other):
-        return other is not None and self.key == other.key
+        return other not in [None, []] and self.key == other.key
 
     def __hash__(self):
         return hash(repr(self))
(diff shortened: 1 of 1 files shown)


Hellowlol:
This needs to be checked carefully. Have you checked that this works as intended manually? There some things that are lists that isn’t populated before a reload is done. Ex actors will only be 3 even if there are more actor.

jjlawren:
It works well for my use case where the attributes listed above disappear during a refresh. This change should still allow the list to be updated if the content changes, but not replaced with an empty list. If there are situations where it would be valid to remove all populated items from a list attribute, this would break it. However, I can't think of any actual examples where this is expected.

jjlawren:
I've been testing this change for a few days and it makes accessing the list attributes very reliable. I haven't encountered any obvious regressions.

PR #601, opened 2020-11-16
Add ability to browse and walk the Plex server system file directories

JonnyWong16:
## Description

Add the ability to browse the Plex server system file directories through the Plex API. This uses the `/services/browse` endpoint when browsing to add a folder to a library.

browse

#### `plexapi.server.PlexServer.browse(path)`

- Browse the system file path using the Plex API.

#### Example

```python

# List all folders and files in a single path
print(plex.browse('/media/Movies'))
```

#### `plexapi.server.PlexServer.walk(path)`

- Walk the system file tree using the Plex API similar to `os.walk`.

#### Example

```py

# Walk through all folders and files in the library locations
for location in plex.library.section('Movies').locations:
    for root, folders, files in plex.walk(location):
        print(root, folders, files)
```

## Type of change

## Checklist:

### DIFF
--- a/tests/test_server.py
+++ b/tests/test_server.py
@@ -287,6 +287,15 @@ def test_server_downloadDatabases(tmpdir, plex):
     assert len(tmpdir.listdir()) > 1
 
 
+def test_server_browse(plex, movies):
+    # browse root
+    paths = plex.browse()
+    assert len(paths)
+    # browse of the files of the movie lib.
+    for path, paths, files in plex.walk(movies.locations[0]):
+        assert len(files)
+
+
 def test_server_allowMediaDeletion(account):
     plex = PlexServer(utils.SERVER_BASEURL, account.authenticationToken)
     # Check server current allowMediaDeletion setting
--- a/tests/test_server.py
+++ b/tests/test_server.py
@@ -288,12 +288,20 @@ def test_server_downloadDatabases(tmpdir, plex):
 
 
 def test_server_browse(plex, movies):
+    movies_path = movies.locations[0]
     # browse root
     paths = plex.browse()
     assert len(paths)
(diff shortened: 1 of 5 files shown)


Hellowlol:
Looks very nice!

tests/test_server.py
```diff
@@ -287,6 +287,15 @@ def test_server_downloadDatabases(tmpdir, plex):
     assert len(tmpdir.listdir()) > 1
 
 
+def test_server_browse(plex, movies):
+    # browse root
+    paths = plex.browse()
+    assert len(paths)
+    # browse of the files of the movie lib.
+    for path, paths, files in plex.walk(movies.locations[0]):
+        assert len(files)
```
blacktwin, line 296: to account for subfolders
`assert len(files) if files else (path and len(paths))`

fcassirer:
Hello. Apologies in advance as I am not familiar with the Plex API capabilities for access restriction, ACLS etc, assuming there is some ability to limit scope via the token.  A capability such as this would seem ripe for mischief if not carefully sandboxed.  Are there security considerations that should be looked at here?

Hellowlol:
@fcassirer I havn’t any done consideration regarding the security. We are only exposing the this for a api. We authenticate using plex own api

tests/test_server.py
Hellowlol: The reason why i only wanted to test for the files is because i don't know what file structure they have choosen.

JonnyWong16: This will fail if there are no files in the root of the movie library folder (i.e. `/media/Movies`, but all files are in subfolders `/media/Movies/Big Buck Bunny(2008)/Big Buck Bunny (2008).mkv`).

PR #590, opened 2020-10-23
Update to MediaPartStream and inheriting classes

blacktwin:
## Description

Clean up of MediaPartStream and it's inheriting classes (video, audio, subtitle).
Moving common attribs among video, audio, and subtitle streams into parent MediaPartStream.
Removal of no longer present attribs from video, audio, subtitle, and MediaPartStream.

Addresses #536

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
Technically not a bug fix, but code clean up.

PR #607, opened 2020-11-21
Add ability to customize XML includes when reloading object

### LINKED ISSUE #603: Make Plex XML "include" parameters configurable
**Describe the issue**
The `_include` parameters when retrieving XML data from Plex should be configurable. It would allow the user to save resources and speed up the response by only retrieving the XML tags/fields that they need.

Example hardcoded `_include` parameters for `~plexapi.video.Movie`:
 https://github.com/pkkid/python-plexapi/blob/163d94d2492e976841f4d7b011cbdef1d849e53e/plexapi/video.py#L289-L291

This is especially useful for the `checkFiles=1` parameter since this causes the Plex server to hit the storage to check if a file exists (`~plexapi.media.MediaPart.exists`) and is accessible (`~plexapi.media.MediaPart.accessible`) which may timeout if cloud storage is being used for the Plex server.

**Code snipppets**
Example of an unnecessary "heavy" reload.
```py

for movie in plex.library.section("Movies").all():
    # This will trigger an automatic reload() of every movie with checkFiles=1
    # even though the exists/accessible values are not needed
    print(movie.subtitleStreams())
```

**Expected behavior**
Example of a possible "lighter" reload.
```py

for movie in plex.library.section("Movies").all():
    # Manual reload() without checkFiles
    movie.reload(checkFiles=False)
    print(movie.subtitleStreams())
```

**Additional context**
Some other XML includes that are available but not shown when viewing the XML through Plex.
  * `includeFields=thumbBlurHash` adds the `thumbBlurHash` field to the XML (used by Plexamp)
  * `includeFields=artBlurHash` adds the `artBlurHash` field to the XML (used by Plexamp)

Reference: [BlurHash](https://blurha.sh/).

JonnyWong16:
## Description

Add ability to customize XML includes when reloading a `PlexObject`.

* The default `_details_key` always has all the includes.
* A partial reload with includes disabled (e.g. `reload(checkFiles=False)`) will remain as a partial object.
* A full `reload()` will have all includes so it will be a full object (existing behaviour maintained).

The only potential breaking change is objects that used to have a blank `_details_key = ''` now have a populated `_details_key = key`. I don't think this should cause any issues because the `_details_key` is only used when reloading and it checks for the `_details_key or key` so it makes no difference if they are now equal.

Fixes #603
Conflicts with #590 (the `_includes` parameters were refactored)
Requires #606 (`isFullObject()` needs to check the `_details_key` instead of `key`)

~~TODO: Tests need to be added for reloading with includes disabled.~~

## Type of change

## Checklist:


plexapi/base.py
```diff
@@ -89,6 +89,22 @@ def _buildItemOrNone(self, elem, cls=None, initpath=None):
...
 
+    def _buildDetailsKey(self, **kwargs):
+        """ Builds the details key with the XML include parameters.
+            All parameters are included by default with the option to override each parameter
+            or disable each parameter individually by setting it to False or 0.
+        """
+        details_key = self.key
+        if hasattr(self, '_includes'):
+            includes = {}
+            for k, v in self._includes.items():
```
Hellowlol, line 100: This should probably be sorted, tests have broken because of the order before.

Hellowlol:
Looks good! We only need tests to verify that the behavior is as expected and that we don’t break anything.

plexapi/base.py
JonnyWong16: I think you want `urlencode(sorted(includes.items()), doseq=True)` below instead. It doesn't matter if the dict is unsorted, but you want URL parameters in the `details_key` to be in the same order.

Hellowlol:
Nice. I noticed some tests that failed because of the pr.

JonnyWong16:
The test needs to be re-run. It looks like my change didn't get pulled into the test.

Line 25:
https://github.com/pkkid/python-plexapi/blob/387ab170723c8a733766cbf72098bc270095a3ad/plexapi/settings.py#L25

But the test says line 24:
https://travis-ci.org/github/pkkid/python-plexapi/jobs/745126871#L886

Hellowlol:
Nice work! 👍 the only thing I think we should improve the the doc string of the kwargs. I think we should mention why this can be usefull as I have algae questions about speed and reloads before.

I’m gonna leave this open for a day so others have had the time to chip in before I merge

JonnyWong16:
Refactored the includes to `PlexPartialObject` so that they are used by Audio, Video, Photoalbum, Photo, Collections, and Playlist.

Doc string for `reload()` is updated.

Hellowlol:
Is the include the same for all the classes?

JonnyWong16:
Most of them are the same when you view them through Plex Web. Some might be extra (e.g. for Playlists) but it doesn't affect the response from the server.

PR #612, opened 2020-12-03
Github Actions followup

jjlawren:
~~The `$default-branch` key didn't seem to be working properly, so hardcoding to `master`.~~

Also updated build badge.


jjlawren:
Since this seems to be triggering and running the CI properly now, I'm suggesting to disable Travis so PRs aren't stuck with pending checks for hours.

PR #615, opened 2020-12-03
Add codecov configuration

jjlawren:
Set some thresholds, skip automated comments to PRs.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -205,10 +205,18 @@ jobs:
 
 
   coverage:
-    name: Process test coverage
+    name: Process test coverage (${{ matrix.plex }})
     runs-on: ubuntu-latest
     needs: pytest
     if: always()
+    strategy:
+      matrix:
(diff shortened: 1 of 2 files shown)


Hellowlol:
I like the automated message as i quickly can see of we miss anything

jjlawren:
This should still enforce some thresholds but won't put the (large) comment in every PR. It feels a bit obtrusive to me but I'm not against keeping it if you see value.

jjlawren:
Unfortunately the PR coverage is always going to drop comparing against `master` since it only runs the unclaimed tests. Not sure of a clean way to address that.

JonnyWong16:
We can split up the coverage reports to `unclaimed` and `claimed`.

See workflow here:
https://github.com/JonnyWong16/python-plexapi/actions/runs/424907307

I don't think anything needs to be changed for the codecov pull requests comments? Only unclaimed reports will be uploaded for pull requests so it should only comment unclaimed coverage.

PR #620, opened 2020-12-07
Fix sync tests when using a token

JonnyWong16:
## Description

Sync tests could fail because of conflicting parallel CI tests using the same token. When a token is created it is associated with a specific `X-Plex-Client-Identifier`. Using the token with different `X-Plex-` headers will update the device attributes at `https://plex.tv/devices.xml` _except_ for the `clientIdentifier`. This causes issues because one CI could be running main tests which sets the header `X-Plex-Provides=controller` while another parallel CI running sync test sets the header `X-Plex-Provides=sync-target`. Depending on which API requests are being made at the time the sync tests could fail because it will suddenly be missing `sync-target` for the device.

This PR:
* Updates `MyPlexPinLogin` to use `https://plex.tv/api/v2` endpoint.
* Adds helper function `createMyPlexDevice` that uses `MyPlexPinLogin` to create a new device on the Plex account.
* Updates sync tests to create a separate sync device with a unique `clientIdentifier`. This gives the sync tests it's own sync device which won't conflict with the device associated with the token.

See workflow run here: https://github.com/JonnyWong16/python-plexapi/actions/runs/404885213

## Type of change

## Checklist:

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -176,11 +176,6 @@ jobs:
 
     - name: Sync tests with ${{ matrix.plex }} server
       if: matrix.plex == 'claimed'
-      env:
-        PLEXAPI_HEADER_PROVIDES: 'controller,sync-target'
-        PLEXAPI_HEADER_PLATFORM: iOS
-        PLEXAPI_HEADER_PLATFORM_VERSION: 11.4.1
-        PLEXAPI_HEADER_DEVICE: iPhone
       run: |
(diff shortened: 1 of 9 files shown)


Hellowlol:
Thanks! Should the test run a claimed server or am it missing something?

jjlawren:
> Thanks! Should the test run a claimed server or am it missing something?

Yes, a claimed server with an account that also has a valid Plex Pass.

PR #629, opened 2020-12-17
Added missing support for /transcode/sessions

red4dj:
I don't think there's an official documented transcode session object but it works properly.


Hellowlol:
Looks like a simple PR. Im holding off on this to a test has been added.

JonnyWong16:
Doc strings and tests added.

PR #632, opened 2020-12-26
Moved split to PlexPartialObject

darthShadow:
Moved split to PlexPartialObject so it's available for the Show object instead of the Episode object

### DIFF
--- a/plexapi/base.py
+++ b/plexapi/base.py
@@ -656,6 +656,11 @@ class PlexPartialObject(PlexObject):
         data = key + '?' + urlencode(params)
         self._server.query(data, method=self._server._session.put)
 
+    def split(self):
+        """Split a duplicate."""
+        key = '/library/metadata/%s/split' % self.ratingKey
+        return self._server.query(key, method=self._server._session.put)
+
     # The photo tag cant be built atm. TODO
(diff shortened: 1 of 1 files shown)


Hellowlol:
Isnt it possible to spilt a episode? I have to look into this

JonnyWong16:
Split apart is only available for movies, tv shows, and artists.
https://support.plex.tv/articles/201018248-merge-or-split-items/

I *think* it might be available for albums? I'm not 100% sure.

darthShadow:
I can confirm that it's available for Shows and Movies and not for Episodes as per the docs.

Can't check for Artists/Albums unfortunately but since they both inherit from the same PlexPartialObject Class, it shouldn't be a problem, even if they do support it.

Hellowlol:
I don't want this added to PlexPartialObject since this class is used for other things that don't support split. The option is to just add this to every class that's need it to create a Mixin.

Hellowlol:
@JonnyWong16 Does the same rule apply for merge?

jjlawren:
@Hellowlol to clarify you'd like to close this PR and use the approach in #639 instead?

Hellowlol:
Yes

PR #630, opened 2020-12-24
Update media type attributes

JonnyWong16:
## Description

This PR is mostly for bringing forward the attribute changes I made to the media type object from my `3.6.0-tautulli` fork. It also unifies the methods between all the media type objects.

Attribute changes:
* Refactored/added new/updated doc strings for `Audio`, `Artist`, `Album`, and `Track` attributes.
* Refactored/added new/updated doc strings for `Video`, `Movie`, `Show`, `Season`, `Episode`, and `Clip` attributes.
* Added new/updated doc strings for `Photoalbum`, and `Photo` attributes.
* Updated doc strings for `Collection` attributes.
* Updated doc strings for `Playlist` attributes.

New objects:
* Added `media.Style` tag object.

New methods:
* `Track.locations` common property to get the `media.Part` locations of the track.
* `Clip.locations` common property to get the `media.Part` locations of the clip.
* `Photo.locations` common property to get the `media.Part` locations of the photo.
* `Photoalbum.clip(title)` method to return a clip in a photo album matching the `title`.
* `Photoalbum.clips()` method to return all clips in a photo album.
* `Photoalbum.get()` alias to `Photoalbum.photo()`.
* `Collection.item(title)` method to return an item in a collection matching the `title`.
* `Collection.get()` alias to `Collection.item()`.
* `Playlist.item(title)` method to return an item in a playlist matching the `title`.
* `Playlist.get()` alias to `Playlist.item()`.

Method changes:
* Separate `title` and `season` parameter in `Show.season()`.
* New `album` name and `track` number parameters in `Artist.track()`.
* New `track` number parameter in `Album.track()`.

Breaking changes:
* Some `index`, `parentIndex`, `grandparentIndex` and `ratingKey`, `parentRatingKey`, `grandparentRatingKey` attributes were type `str` and have now all been casted to `int` for consistency.
* The `collections` attribute is removed from `Episode` since you cannot add an episode to a collection.
* All `titleSort` attributes now default to `title` instead of returning a blank string.
* ~~The `title` parameter in `Show.season()` used to accept an integer for the season number which is now changed to the `season` parameter.~~ Kept check for `isinstance(title, int)` for backwards compatibility.

## Type of change

Please delete options that are not relevant.

- [x] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] This change requires a documentation update

## Checklist:

- [x] My code follows the style guidelines of this project
[...]

PR #633, opened 2020-12-27
Fix 'Clip' object has no attribute 'media' error for Clip.download() method

kolsys:
Fixing `AttributeError: 'Clip' object has no attribute 'media'` while calling Object.download method. It's useful when loading video objects from Photo library.


JonnyWong16:
`media` attribute added in #630.

PR #634, opened 2020-12-30
Use common session and increase Plex account timeout in tests

Hellowlol:


jjlawren:
Do you remember what this was in response to?

Hellowlol:
Timeout error regarding some myplex tests

PR #636, opened 2021-01-03
Add Plex dashboard bandwidth and resources

JonnyWong16:
## Description

Adds support for the bandwidth and resources data on the Plex dashboard.

#### New objects:
* `server.SystemAccount` is an account associated with the server. (Updated existing object)
* `server.SystemDevice` is a device associated with the server.
* `server.StatisticsBandwidth` is the bandwidth data for a single account / device / timestamp combination.
* `server.StatisticsResources` is the resource data for a single timestamp.

#### New methods:
* `PlexServer.systemAccounts()` returns a list of accounts associated with the Plex server. (Updated existing method)
* `PlexServer.systemDevices()` returns a list of devices associated with the Plex server.
* `PlexServer.bandwidth()` will return a list of the dashboard bandwidth data. The `timespan` parameter will bin the bandwidth data and `**kwargs` can be used to filter the bandwidth data.
* `PlexServer.resources()` will return a list of the dashboard CPU and RAM usage data. Only live data (1 second intervals) is available.

#### Other changes:
* `MyPlexAccount.id` casted to an `int` instead of a `str`.

#### Example:
```py

# Filter bandwidth data for December 2020 and later and more than 1 GB used.
filters = {
    'at>': datetime(2020, 12, 1),
    'bytes>': 1024**3
}

# Retrieve bandwidth data in one day timespans.
bandwidthData = plex.bandwidth(timespan='days', **filters)

# Print out bandwidth usage for each account and device combination
for bandwidth in sorted(bandwidthData, key=lambda x: x.at):
    account = bandwidth.account()
    device = bandwidth.device()
    gigabytes = round(bandwidth.bytes / 1024**3, 3)
    local = 'local' if bandwidth.lan else 'remote'
    date = bandwidth.at.strftime('%Y-%m-%d')
    print('%s used %s GB of %s bandwidth on %s from %s'
          % (account.name, gigabytes, local, date, device.name))
```

Partially addresses #356

TODO: Add some tests. I don't know if bandwidth tests will work since the test server won't have any history. Maybe just check that the dashboard returns *something* using the live `seconds` interval.

## Type of change

## Checklist:

### DIFF
--- a/plexapi/server.py
+++ b/plexapi/server.py
@@ -107,6 +107,8 @@ class PlexServer(PlexObject):
         self._library = None   # cached library
         self._settings = None   # cached settings
         self._myPlexAccount = None   # cached myPlexAccount
+        self._systemAccounts = None # cached list of SystemAccount
+        self._systemDevices = None # cached list of SystemDevice
         data = self.query(self.key, timeout=timeout)
         super(PlexServer, self).__init__(self, data, self.key)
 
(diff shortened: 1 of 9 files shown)


plexapi/server.py
```diff
@@ -607,6 +616,105 @@ def _allowMediaDeletion(self, toggle=False):
...
+            elif key.startswith('bytes') or key == 'lan':
+                value = cast(int, value)
+            elif key == 'accountID':
+                if value == self.myPlexAccount().id:
+                    value = 1  # The admin account is accountID=1
+            params[key] = value
+
+        key = '/statistics/bandwidth?%s' % urlencode(params)
+        data = self.query(key)
+        return self.findItems(data, StatisticsBandwidth)
```
Hellowlol, line 709: Is query and findItems required where for some reason or would it be possible to use fetchItems?

Hellowlol:
Nice job! About the testing, would it be possible to just add some \data in tests/payloads.py so we have something to test against?

JonnyWong16:
So I didn't need to mock bandwidth, but I had to mock the resources. I don't know why the test server can't read CPU/RAM usage.

Passing tests are here. (Note that I modified the CI to only run the server tests in this example.)
https://github.com/JonnyWong16/python-plexapi/actions/runs/459798422

PR #658, opened 2021-02-06
Add ability to upload theme music

ReenigneArcher:
Added theme endpoints
Fixed comments in art endpoints.

### DIFF
--- a/plexapi/audio.py
+++ b/plexapi/audio.py
@@ -223,7 +223,25 @@ class Artist(Audio):
             for track in album.tracks():
                 filepaths += track.download(savepath, keep_original_name, **kwargs)
         return filepaths
+    
+    def themes(self):
+        """ Returns list of available theme objects. :class:`~plexapi.media.Theme`. """
 
+        return self.fetchItems('%s/themes' % self.key)
+
+    def uploadTheme(self, url=None, filepath=None):
+        """ Upload theme from url or filepath. :class:`~plexapi.media.Theme` to :class:`~plexapi.video.Video`. """
+        if url:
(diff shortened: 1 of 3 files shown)


Hellowlol:
This class supports many things? Does everything has a theme?

ReenigneArcher:
> 
> 
> This class supports many things Does everything has a theme?

According to this old documentation, only movie, tv show, and artist (not tv episode or tv season)

https://web.archive.org/web/20150113085312/http://dev.plexapp.com/docs/agents/models.html

Where should I put it to limit what items can have a theme?

ReenigneArcher:
Think I found where... videos.py

ReenigneArcher:
Sorry, I don't understand why the checks are failing...

JonnyWong16:
Themes should be moved to a mixin similar to #648.

blacktwin:
This isn't something that we can do in the UI currently and the way you'd undo it is potentially destructive to the item. [From my experience of playing this feature](https://forums.plex.tv/t/managing-theme-music-posters-and-art-with-the-webui-and-api/557292), you'd have to dance the item away to clear out the themes. I'd assume when danced, because the item is given a new ratingKey, all it's history is gone. It's why I didn't create a PR then and posted the feature request instead. 

Maybe add a disclaimer of using the feature or guidance on what to do (dance) if you regret your decision. For Shows libraries, where there is a toggle for the library, you can turn off themes. For non Show libraries you can't. So unless you upload a silent audio file you're stuck with a theme on that item or you dance.

ReenigneArcher:
Closing as #879 replaces this and I longer have access to the repo I made these changes in.

PR #659, opened 2021-02-07
Don't refresh metadata after editing tags

JonnyWong16:
Just two additional minor commits that should have been included in #649.

Also workaround #660 for reloading object in mixins tag test.


Hellowlol:
Good stuff. We should never refresh or reload when commands like adding label as this will slow down all batch commands.

PR #668, opened 2021-02-15
Fix gdm.GDM.find_by_content_type()

Montellese:
While trying to integrate `plexapi.gdm.GDM.find_by_content_type()` into my service I noticed that it always throws an error because it tries to extract `Content_Type` instead of `Content-Type` from the received data. The last commit of this PR fixes that.

I added two additional commits:
* the first commit removes the unused attribute `last_scan`
* the second commit adds missing documentation for the `entries` attribute


jjlawren (approved):
Thanks, looks good!

PR #692, opened 2021-03-11
Handle missing key when PlayQueue passed to playMedia

jjlawren:
If a `PlayQueue` object is passed to `PlexClient.playMedia()`, the generated `key` argument used in the resulting call is always `None`. This uses the currently selected item from the `PlayQueue` in that case.

### DIFF
--- a/plexapi/client.py
+++ b/plexapi/client.py
@@ -488,23 +488,16 @@ class PlexClient(PlexObject):
         if mediatype == "audio":
             mediatype = "music"
 
-        if isinstance(media, PlayQueue):
-            playQueueID = media.playQueueID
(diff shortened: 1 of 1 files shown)


jjlawren:
An alternative that's less explicit but far simpler: https://github.com/jjlawren/python-plexapi/commit/bb7b959c16bcb2931199bbaa28ad672ededdf0cd

jjlawren:
The `key` value has been sending `None` to clients for years when a premade playqueue is passed in, but most seem to handle it gracefully and play anyway. Others like the 3rd party Samsung TV "Orca" client do not and throw up an error.

PR #698, opened 2021-03-18
Added basic support for Watch Together feature

nwithan8:
Get live data on the Rooms for the new Watch Together feature.

Haven't discovered any control endpoints yet, so this is read-only

(Proper PR this time)


JonnyWong16:
Please use the PR template and checklist.

PR #704, opened 2021-03-21
Add deviceID to Playable for history entries

glensc:
This is needed to identify devices from play history:

```python
    history = movie.history()
    for h in history:
        print("- Watched: %s (%s) -> %s,%s" % (h.viewedAt, h.lastViewedAt, h.accountID, h.deviceID))
```

### DIFF
--- a/plexapi/base.py
+++ b/plexapi/base.py
@@ -538,6 +538,7 @@ class Playable(object):
         self.session = self.findItems(data, etag='Session')                         # session
         self.viewedAt = utils.toDatetime(data.attrib.get('viewedAt'))               # history
         self.accountID = utils.cast(int, data.attrib.get('accountID'))              # history
+        self.deviceID = utils.cast(int, data.attrib.get('deviceID'))                # history
         self.playlistItemID = utils.cast(int, data.attrib.get('playlistItemID'))    # playlist
         self.playQueueItemID = utils.cast(int, data.attrib.get('playQueueItemID'))  # playqueue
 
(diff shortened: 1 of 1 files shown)


glensc:
Also need some help, how to find Device based on that Id?

There's `plex.systemDevices()`, but it returns all devices, not the specific one I'm interested of.

Also, the offset is off by one when accessing as index, so need to use lookup;

I've come up with this:
```python
    devices = plex.systemDevices()
    history = movie.history()
    for h in history:
        # device = devices[h.deviceID] <- devices is not indexed by deviceId
        device = next(x for x in devices if x.id == h.deviceID)
```

plexapi/base.py
```diff
@@ -538,6 +538,7 @@ def _loadData(self, data):
         self.session = self.findItems(data, etag='Session')                         # session
         self.viewedAt = utils.toDatetime(data.attrib.get('viewedAt'))               # history
         self.accountID = utils.cast(int, data.attrib.get('accountID'))              # history
+        self.deviceID = utils.cast(int, data.attrib.get('deviceID'))                # history
```
JonnyWong16, line 541: Could you also add `accountID` and `deviceID` to the doc string?

JonnyWong16 (changes requested):

JonnyWong16:
Using `next()` will work.

I added convenience methods to retrieve the account or device by ID in #703.

plexapi/base.py
glensc: f6596a03cea3e69a1e1e93aa29141fe261d25720

glensc:
> Using `next()` will work.
> 
> I added convenience methods to retrieve the account or device by ID in #703. The methods just use `next()` the same way.

Maybe (also) make the devices indeed by deviceId?

JonnyWong16:
> Maybe (also) make the devices indeed by deviceId?

I don't know what you mean.

JonnyWong16 (approved):

Issue #706, opened 2021-03-21
Add API to override Platform


glensc:
**Describe the issue**

There's currently no programmatic way to override Platform and Client name.

Hacks like overwriting `X_PLEX_PLATFORM` / `X_PLEX_DEVICE_NAME` do not work without more internal symbols hacking, because those values are copied elsewhere:
- https://github.com/pkkid/python-plexapi/blob/1429867ae5b0bb92bc9db373f7dec36bc035b686/plexapi/__init__.py#L23-L32

Setting environment variables also does not work, as that needs to be done before the `import plexapi` is called.

**Code snippets**

```python
from os import getenv, environ
import plexapi

        plex_config = {
            'PLEXAPI_HEADER_PLATFORM': 'PlexTraktSync',
            'PLEXAPI_HEADER_DEVICE_NAME': 'PlexTraktSync',
        }
        plexapi.X_PLEX_DEVICE_NAME = 'PlexTraktSync'
        for k, v in plex_config.items():
            environ[k] = v
```

**Expected behavior**

```python
    plex = PlexServer(url, token, device_platform='PlexTraktSync')
```

**Additional context**

Trying to make changes identifiable to be able to skip own changes. I.e skip changes identified by named device:
- https://github.com/Taxel/PlexTraktSync/discussions/146#discussioncomment-507487

Hellowlol:
I don’t understand the problem. Your describing different ways to set this that does work. You have to do it in the correct order. 

I also fail to see how the issue you linked to is relevant.

glensc:
@Hellowlol those examples are all various hacks. Therefore I request a programmatic way to set the deviceId in the plex instance, not the global static module internal variable. Hence creating this issue for enchancement. Having some trickery done before importing a module trashes up project coding style. Python coding style is that you put `import`'s in the beginning of the file, and then in `main()` you execute the code.

Besides messing up style, it's very fragile, if in case somebody else loads the module before the header overwrites the code, your initialization will fail.

Hellowlol:
I can’t speak for others I’m not gonna do this at all.

There are 3 ways to do this. Patching, environmental variables and using a config file.

Your assuming that PlexServer is the main entry point. That might be true for simple scripts, but what about all the others? Should we add device_platform to all other classes that ppl might, such as the cloud classes? I assume you see my point, as this will be very had to maintain.

With that being said, if you can come up with a simpler method then we are currently using that is easy to maintain I’ll happily merge that.

[...]

PR #707, opened 2021-03-21
Initialize headers from config in reset_base_headers()

glensc:
WIP: https://github.com/pkkid/python-plexapi/issues/706


plexapi/client.py
```diff
@@ -126,7 +128,7 @@ def _loadData(self, data):
 
     def _headers(self, **kwargs):
         """ Returns a dict of all default headers for Client requests. """
-        headers = BASE_HEADERS
+        headers = self._base_headers
         if self._token:
             headers['X-Plex-Token'] = self._token
         headers.update(kwargs)
```
glensc, line 134: Here's something that the previous code did not consider:

```python
dict1 = dict2
dict1.update(kwargs)
```

this updates `dict2` as well, i.e any such call would have modified `BASE_HEADERS` which would interfere with other calls, even from different classes.

likely never to be discovered as they all fill `X-Plex-Token` there with the same value.

this can be easily fixed by calling `headers = resrt_base_headers()` instead of assignment. so which way to go? optimize for using self._base_headers from constructor, or fix for correctness and use `reset_base_headers()` call in `_headers()`?

Hellowlol: It’s just a bug. We use copy other places.

Don’t break anything and use camelCase

glensc: sorry, camelCase for what?

glensc:
OK, here's some idea that could work. Don't mind the PR title and body, will update them once the PR comes useful.

So, WDYT, is this acceptable?

plexapi/config.py
```diff
@@ -42,24 +42,46 @@ def get(self, key, default=None, cast=None):
...
     return {
-        'X-Plex-Platform': plexapi.X_PLEX_PLATFORM,
-        'X-Plex-Platform-Version': plexapi.X_PLEX_PLATFORM_VERSION,
-        'X-Plex-Provides': plexapi.X_PLEX_PROVIDES,
-        'X-Plex-Product': plexapi.X_PLEX_PRODUCT,
-        'X-Plex-Version': plexapi.X_PLEX_VERSION,
-        'X-Plex-Device': plexapi.X_PLEX_DEVICE,
-        'X-Plex-Device-Name': plexapi.X_PLEX_DEVICE_NAME,
-        'X-Plex-Client-Identifier': plexapi.X_PLEX_IDENTIFIER,
+        'X-Plex-Platform': CONFIG.get('header.platorm', CONFIG.get('header.platform')),
```
glensc, line 78: Maybe it's about time to drop support for the `'header.platorm'` typo?

glensc: it was fixed in 2018, matching 3.2.0 version:
- https://github.com/pkkid/python-plexapi/pull/291

glensc:
How about this approach? If needed, I could extract to smaller changes.

Hellowlol:
I still don’t see the point. IMO you need to add a example on why this is better.  Fix your flake errors, run the test suit locally, including the tests for the client and squash the commits.

I don’t like the name reset_default_headers as isn’t anymore, is the headers is in config file (is was a bad name in the first place) Why is the headers a local variable the class? Why not use it directly?

plexapi/config.py
```diff
@@ -42,24 +42,46 @@ def get(self, key, default=None, cast=None):
...
         config = defaultdict(dict)
+        config.update(self._defaults())
         for section in self._sections:
             for name, value in self._sections[section].items():
                 if name != '__name__':
                     config[section.lower()][name.lower()] = value
         return dict(config)
 
+    def _defaults(self):
+        from uuid import getnode
```
Hellowlol, line 53: Don’t in-line std libs.

plexapi/config.py
```diff
@@ -42,24 +42,46 @@ def get(self, key, default=None, cast=None):
...
+        from platform import uname
+        from plexapi import PROJECT, VERSION
+
+        platform_name, device_name, platform_version = uname()[0:3]
```
Hellowlol, line 57: Fix the slicing.

glensc: Explain fix how? fix what? what is wrong with the current code?

glensc:
> Fix your flake errors, run the test suit locally, including the tests for the client and squash the commits.

I don't see point fixing tests until it's certain this is worth finishing! That's why I asked about the changes, not the style! Also, do note the PR is in a Draft state.

glensc:
> I don’t like the name reset_default_headers as isn’t anymore, is the headers is in config file (it was a bad name in the first place)

I used an existing name, please suggest a better name not only complain it's bad.

glensc:
> Why is the headers a local variable the class? Why not use it directly?

I think I explained it in an issue this PR is referring to (https://github.com/pkkid/python-plexapi/issues/706). to make the headers instance-based, not global. so the `self._headers` is now copied into the instance. is that `headers` you are referring to?

since Config object may be modified runtime (after `__init__`) using copy initialized in `__init__.py` is the original problem. 

ps: since Github has no threading in comment mode, submit a review as code comments, so can reply in a thread and resolve disucssions.

plexapi/config.py
glensc: why is the current slicing wrong. how is the correct way?

PR #722, opened 2021-04-06
fix for advanced settings that dont use ints

meisnate12:
## Description

I was trying to change the Ratings Source Advance Setting of a Library like so:

```python
ps = PlexServer("http://192.168.1.144:32400", "notmyactualtoken", timeout=600)
plex = next((s for s in ps.library.sections() if s.title == "New Movies"), None)
plex.editAdvanced(**{"ratingsSource": "themoviedb"})
```
 
and i would get this:

```cmd
  File "C:\Users\meisn\Documents\Python\Plex Meta Manager\test.py", line 184, in <module>
    plex.editAdvanced(**{"ratingsSource": "themoviedb"})
  File "C:\Users\meisn\Documents\Python\Plex Meta Manager\venv\lib\site-packages\plexapi\library.py", line 508, in editAdvanced
    enumValues = [int(x) for x in enums]
  File "C:\Users\meisn\Documents\Python\Plex Meta Manager\venv\lib\site-packages\plexapi\library.py", line 508, in <listcomp>
    enumValues = [int(x) for x in enums]
ValueError: invalid literal for int() with base 10: 'rottentomatoes'
```

From what i can tell this is happening because all the keys from the enumValues of each setting get cast to ints, which works for most settings except ratingSource and one or two others. I just caught the ValueError and then recreated the list without casting it to int and i was able to edit the advance setting. 

## Type of change

## Checklist:

### DIFF
--- a/plexapi/library.py
+++ b/plexapi/library.py
@@ -505,7 +505,10 @@ class LibrarySection(PlexObject):
         for settingID, value in kwargs.items():
             try:
                 enums = idEnums.get(settingID)
-                enumValues = [int(x) for x in enums]
+                try:
+                    enumValues = [int(x) for x in enums]
+                except ValueError:
+                    enumValues = [x for x in enums]
             except TypeError:
                 raise NotFound('%s not found in %s' % (value, list(idEnums.keys())))
             if value in enumValues:
--- a/plexapi/library.py
+++ b/plexapi/library.py
@@ -504,14 +504,10 @@ class LibrarySection(PlexObject):
 
         for settingID, value in kwargs.items():
             try:
-                enums = idEnums.get(settingID)
-                try:
-                    enumValues = [int(x) for x in enums]
-                except ValueError:
-                    enumValues = [x for x in enums]
-            except TypeError:
+                enums = idEnums[settingID]
+            except KeyError:
                 raise NotFound('%s not found in %s' % (value, list(idEnums.keys())))
-            if value in enumValues:
+            if value in enums:
                 data[key % settingID] = value
             else:
                 raise NotFound('%s not found in %s' % (value, enums))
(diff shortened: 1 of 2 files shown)


JonnyWong16:
I think `enumValues` is completely redudnant.
```python
        for settingID, value in kwargs.items():
            try:
                enums = idEnums[settingID]
            except KeyError:
                raise NotFound('%s not found in %s' % (settingID, list(idEnums.keys())))
            if value in enums:
                data[key % settingID] = value
            else:
                raise NotFound('%s not found in %s' % (value, enums))
```

meisnate12:
so i think this was casting to int because of the advanced setting episodeSort. It's enumValues' keys are numerical but stored as strings, so Im assuming that it was in place so they could all be ints.
```python
plex = next((s for s in ps.library.sections() if s.title == "TV Shows"), None)
for s in plex.settings():
    if s.id == "episodeSort":
        print(s, s.enumValues)
```
```
<Setting:episodeSort:-1> {'-1': 'Library default', '0': 'Oldest first', '1': 'Newest first'}
```

JonnyWong16:
That's a bug that should be reported to Plex and fixed on their end. Their API lists `episodeSort` as `type="text"` instead of `int.
```xml
<Setting id="episodeSort" label="Episode sorting" summary="How to sort the episodes for this show." type="text" default="-1" value="-1" hidden="0" advanced="0" group="" enumValues="-1:Library default|0:Oldest first|1:Newest first"/>
```

meisnate12:
ok I removed the redundant variable and updated `mixins.AdvancedSettingsMixin.editAdvanced()`

JonnyWong16:
`episodeSort ` data type bug reported to Plex.

https://forums.plex.tv/t/bug-incorrect-data-type-in-api-for-tv-show-episodesort-setting/711629

PR #729, opened 2021-04-23
Add new collection attributes

JonnyWong16:
## Description

Add new collection attributes related to smart collections. Actual support for creating smart collections to come later.

Seasons, episodes, and tracks can also be added to collections now.

## Type of change

## Checklist:

### DIFF
--- a/plexapi/audio.py
+++ b/plexapi/audio.py
@@ -328,13 +328,14 @@ class Album(Audio, ArtMixin, PosterMixin, UnmatchMatchMixin,
 
 
 @utils.registerPlexObject
-class Track(Audio, Playable, ArtUrlMixin, PosterUrlMixin, MoodMixin):
+class Track(Audio, Playable, ArtUrlMixin, PosterUrlMixin, CollectionMixin, MoodMixin):
     """ Represents a single Track.
 
         Attributes:
             TAG (str): 'Directory'
             TYPE (str): 'track'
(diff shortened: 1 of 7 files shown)


meisnate12:
collectionPublished will cause a ValueError to be thrown whenever you access the collection if anything besides 0 or 1 is stored in the field (which is currently possible)

JonnyWong16:
Don't put random numbers in the setting... Plex has already said that the input box in the UI is an error and they will be removing it.

meisnate12:
gotcha I was testing if the number there had anything to do with whether or not a collection was published on the homepage and just noticed that it caused this error so just figured I'd let you know

JonnyWong16:
It's a bool, so it's only 1 (True) or 0 (False).

meisnate12:
It's just that the average user doesn't know that so if someone just puts something there not knowing any better and then tries to run a script that loads that collection it's going to error every time the script tries to access that collection. While it may be a bool the UI looks like a string and allows a string value to be stored there so I feel at the very least this shouldn't error just trying to load the collection when there is a string.

JonnyWong16:
1. I'm just letting you know that it's a bool, not saying everyone needs to know.
2. Plex has already said they are removing it from the UI.
3. This hasn't been merged yet, so there is no issue with anyone accidentally using it.

PR #731, opened 2021-04-26
add library labels endpoint

meisnate12:
## Description

I added the Labels endpoint (`/library/sections/<section_id>/label`) which returns the new Label object. This gets you every label in that library. I could use some guidance on writing proper tests or if my docstring is incorrect but when tested for me it works.

## Type of change

## Checklist:


meisnate12:
yup seems to be so i had looked at where the new advance filters were calling the label list and couldn't find that endpoint so i added it, but yea `LibrarySection.listFilterChoices(field='label')` worked just the same ill close then

PR #765, opened 2021-05-30
Allow overwriting attributes with None when manually reloading

### LINKED ISSUE #713: Accessing optional attributes results in an unnecessary reload
**Describe the issue**

This is not a bug but a feature / improvement request.

I'm trying to retrieve all items from a library section e.g. all movies. Then I'm collecting all the information available for the retrieved movies which requires a reload for some of the attributes like `Movie.roles` because not all roles are present in the XML data received from the `/library/sections/{id}/all` endpoint. Since I'm not interested in all the extra stuff like `On Deck`, `Related` etc. I wanted to make use of the possibility to customize the used `includeXYZ` in `PlexObject.reload()` as introduced in #607. While the customized call to `PlexObject.reload()` works fine, in the end `plexapi` often ends up performing a full `reload()` anyway because I try to access an attribute of `Movie` or `Video` which is optional in the XML response like `lastViewedAt` (which is only provided if the movie has been viewed at least once). The problem is that for all unviewed movies `Movie.lastViewedAt` will be `None` and the `Movie` object will only be a partial object because I did a partial `reload()` so the logic in `PlexObject.reload()` will perform a full reload in addition to the customized partial `reload()` I executed before.

https://github.com/pkkid/python-plexapi/blob/2bde2344084721d7015167af4762d3bc397fe5d2/plexapi/base.py#L426-L442

Is there any way we can add a list of attributes to the specific media classes like `Video`, `Movie` etc. which contains all attributes which don't require a full `reload()` even if they are `None` or `[]`?

PS: This does not only apply to `lastViewedAt` but also to `collections`, `originalTitle` and others.

**Code snippets**

The following code snippet first performs a customized `reload()` and then internally a full `reload()` if `lastViewedAt` is not set because the video has never been played / watched:
```python
item.reload(includeOnDeck=False, includeRelated=False, includeReviews=False)
print(f"{item.title} last viewed at {item.lastViewedAt}")
```

**Expected behavior**

Only the customized `reload()` actually reloads the item while accessing `lastViewedAt` doesn't perform a full `reload()`.

### LINKED ISSUE #770: .reload() fails to show updated metadata when 'studio.value' is set to '' (ie, removed)
I was trying to remove (not change) the studio for a given movie, and I noticed something odd.  When you pass in `**{'studio.value': None [...]}` as an arg to `edit()`, `urllib.parse.urlencode` encodes the `None` as a string: `'None'`, setting the studio name to `'None'` rather than removing it.  

```python
print(movie.studio)  # 'FakeStudio'
k = {'studio.value': None, 'studio.locked':0}
movie.edit(**k)
movie.reload()
print(movie.studio)  # 'None'  
# NOT None, but the string 'None'
```

This is kind of annoying, but not horrible (though maybe should be documented).  

I found that passing in `**{'studio.value': '' [...]}` will work instead.

**HOWEVER**, it took me _ages_ to realize this, because of the bug.

```python
print(movie.studio)  # 'FakeStudio'
k = {'studio.value': '', 'studio.locked':0}
movie.edit(**k)
movie.reload()
print(movie.studio)  # 'FakeStudio'
```

**Although the movie in actuality now has its studio removed, reloading the the movie will display whatever studio it had before, as if it hadn't changed**.  This appears to be true only when the 'studio.value' key has a value of `''`---all other strings will display correctly when updated.

JonnyWong16:
## Description

The current reload behaviour prevents overwriting attributes with `None`. This change allows overwriting attributes with `None` when calling a manual `reload()`, but still prevents overwriting when we do the automatic internal reload when accessing a missing attribute.

Other changes:
* Changes `_DONT_RELOAD_FOR_KEYS` and `_DONT_OVERWRITE_SESSION_KEYS` to private variables.
* Adds `USER_DONT_RELOAD_FOR_KEYS` public variable for user configurable keys (Closes #713).

~~TODO: Tests can be updated after rebasing on #763 and #764 which ran into this reloading issue.~~
* ~~#763 - Erasing a collection's `contentRating` resets it back to `None`.~~
* ~~#764 - Resetting an items `userRating` resets it back to `None`.~~

Fixes #770

## Type of change

## Checklist:

### DIFF
--- a/plexapi/base.py
+++ b/plexapi/base.py
@@ -302,7 +302,7 @@ class PlexObject(object):
                 results.append(elem.attrib.get(attr))
         return results
 
-    def reload(self, key=None, _autoReload=False, **kwargs):
+    def reload(self, key=None, **kwargs):
         """ Reload the data for this object from self.key.
 
             Parameters:
@@ -333,6 +333,10 @@ class PlexObject(object):
                     movie.isFullObject()  # Returns True
 
         """
+        return self._reload(key=key, **kwargs)
+
+    def _reload(self, key=None, _autoReload=False, **kwargs):
+        """ Perform the actual reload. """
         details_key = self._buildDetailsKey(**kwargs) if kwargs else self._details_key
         key = key or details_key or self.key
         if not key:
@@ -459,7 +463,7 @@ class PlexPartialObject(PlexObject):
         objname = "%s '%s'" % (clsname, title) if title else clsname
         log.debug("Reloading %s for attr '%s'", objname, attr)
         # Reload and return the value
-        self.reload(_autoReload=True)
+        self._reload(_autoReload=True)
         return super(PlexPartialObject, self).__getattribute__(attr)
 
     def analyze(self):
@@ -701,7 +705,7 @@ class Playable(object):
         key = '/:/progress?key=%s&identifier=com.plexapp.plugins.library&time=%d&state=%s' % (self.ratingKey,
                                                                                               time, state)
         self._server.query(key)
-        self.reload(_autoReload=True)
+        self._reload(_autoReload=True)
 
     def updateTimeline(self, time, state='stopped', duration=None):
         """ Set the timeline progress for this video.
@@ -719,4 +723,4 @@ class Playable(object):
         key = '/:/timeline?ratingKey=%s&key=%s&identifier=com.plexapp.plugins.library&time=%d&state=%s%s'
         key %= (self.ratingKey, self.key, time, state, durationStr)
         self._server.query(key)
-        self.reload(_autoReload=True)
+        self._reload(_autoReload=True)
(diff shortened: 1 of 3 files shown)


plexapi/base.py
```diff
@@ -443,8 +453,9 @@ def __getattribute__(self, attr):
         # Dragons inside.. :-/
         value = super(PlexPartialObject, self).__getattribute__(attr)
         # Check a few cases where we dont want to reload
-        if attr in DONT_RELOAD_FOR_KEYS: return value
-        if attr in DONT_OVERWRITE_SESSION_KEYS: return value
+        if attr in _DONT_RELOAD_FOR_KEYS: return value
+        if attr in _DONT_OVERWRITE_SESSION_KEYS: return value
+        if attr in USER_DONT_RELOAD_FOR_KEYS: return
```
Montellese, line 458: Do you not `return value` here on purpose?

JonnyWong16: Oops I forgot.

PR #834, opened 2021-09-26
Add support for AlertListener receiving errors

glensc:
## Description

Add support for AlertListener receiving errors

Use the NodeJs like `callback(success, error)` signature for distinguishing errors from data.

## Type of change

## Checklist:

### DIFF
--- a/plexapi/alert.py
+++ b/plexapi/alert.py
@@ -72,7 +72,7 @@ class AlertListener(threading.Thread):
             data = json.loads(message)['NotificationContainer']
             log.debug('Alert: %s %s %s', *data)
             if self._callback:
-                self._callback(data)
+                self._callback(data, None)
         except Exception as err:  # pragma: no cover
             log.error('AlertListener Msg Error: %s', err)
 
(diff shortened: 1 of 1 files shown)


glensc:
This is a somewhat breaking change, listeners may receive now `None` as data.

What do you think of this API? or better make the change not breaking change? for example have two callback parameters for the AlertListener constructor?

Hellowlol:
I dont like it. Dumping a exception as a string is kinda useless imo

JonnyWong16:
I don't like the `self._callback(data, None)` and `self._callback(None, err)`. Passing around `None` like that seems very "unpythonic".

If this is implemented, I would prefer having two different callback functions. It's cleaner and isn't a breaking change.

```py
AlertListener(callback, callbackError)
```

The only use case I can think of for having two different callback functions is to log the error message to your own application. The error isn't a Plex error, it's a websocket client error.

Hellowlol:
The user can log what they want inside the callback.

glensc:
I use this event to reset caches, as likely server was restart. as plex sessions watch are dumb incremental numbers. they will conflict if server restart.

Hellowlol:
If you need a robust way to connect to pms using websocket you could use another package. I needed ut using another one as i wanted one that always reconnect when the connection drops.

https://github.com/Hellowlol/bw_plex/blob/master/bw_plex/plex.py#L1308-L1322

glensc:
Replaced by https://github.com/pkkid/python-plexapi/pull/850

PR #835, opened 2021-09-26
Fix Environment typo in bug report template

glensc:


JonnyWong16:
I was actually going to replace the issues template with forms instead but haven't gotten around to it yet.

https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository

See Tautulli for example.

https://github.com/Tautulli/Tautulli/tree/master/.github/ISSUE_TEMPLATE

Issue #827, opened 2021-09-11
Feature Request - Support Edit / Add of Multiple Locations for Library Sections


EarthBoundX5:
**Describe the issue**

The current Python-PlexAPI does not appear to support edit() or add() of libraries with multiple folder locations.  The vanilla Plex API supports this function.  Please add functionality for feature parity.

**Code snippets**

Here is some general pseudo code that may be a way to implement this functionality.

```
if librarylocation.length > 1
x=librarylocation.length
while x != 0
locations = locations + "librarylocation[x]" + "&"
x--
trim locations "&"
```

**Expected behavior**

I would expect to be able to either pass in an array or string of values matching a defined syntax to Python-PlexAPI and it be passed to the Plex API request.

The current Plex API achieves this by simply having multiple location='s in the PUT request URL.  For example:
`location=C:\Users\Administrator\Desktop\TempPlex\Temp1&location=C:\Users\Administrator\Desktop\TempPlex\Temp2&location=C:\Users\Administrator\Desktop\TempPlex\Temp3
`

**Enviroment (please complete the following information):**

 - OS: Windows Server 2019
 - Plex version: 1.24.1.4931
 - Python Version: 3.7.9
 - PlexAPI version: 4.7.0

**Additional context**

I discovered this deficiency while posting in the Support Discord channel and received input and assistance from SwiftPanda.  I think this could be a very useful supported function for Python-PlexAPI, I had planned to use it in tandem with automating library adds; but will be using PlexAPI direct now, so unless I have issues, there is no urgency from me on this request.

Additionally, worth noting there is a misspelling above in the GitHub issue template.  Enviroment is spelled Environment.

EarthBoundX5:
In case it helps, I've posted my solution to this using the base URL + PowerShell here:
https://www.reddit.com/r/PleX/comments/plbd27/scripted_adding_of_folders_to_libraries/

JonnyWong16:
This was added in version [4.9.0](https://github.com/pkkid/python-plexapi/releases/tag/4.9.0).

PR #837, opened 2021-09-27
Library editing updates

blacktwin:
## Description

Adding convenience methods to add and remove location(s) (paths) to a library. Updating library `add()` method to allow for muliple locations on creation of library.

Fixes or addresses #827

## Example

```python
plex = PlexServer(PLEX_URL, PLEX_TOKEN, session=sess)

locations = ["D:\\plexlibraries\\shows",
             "D:\\plexlibraries\\shows_optimized"]

plex.library.add(name='Test Library', type='show', agent='tv.plex.agents.series',
                      scanner='Plex TV Series', location=locations, language='U')
					  
testLibrary = plex.library.section('Test Library')
testLibrary.removeLocations(locations[0])
testLibrary.addLocations(locations[0])
```

## Type of change

## Checklist:


blacktwin:
TODO: Add tests

plexapi/library.py
```diff
@@ -294,8 +294,13 @@ def add(self, name='', type='', agent='', scanner='', location='', language='en'
                   40:South Africa, 41:Spain, 42:Sweden, 43:Switzerland, 44:Taiwan, 45:Trinidad,
                   46:United Kingdom, 47:United States, 48:Uruguay, 49:Venezuela.
         """
-        part = '/library/sections?name=%s&type=%s&agent=%s&scanner=%s&language=%s&location=%s' % (
-            quote_plus(name), type, agent, quote_plus(scanner), language, quote_plus(location))  # noqa E126
+        if isinstance(location, str):
+            if not utils.pathExist(location):
```
JonnyWong16, line 298: Using `os.path.isdir(path)` won't work if plexapi is on a different computer than the Plex server.

Use `self._server.browse(path)` instead which mimics Plex Web. I don't know if `server.browse(path)` returns any type of error if the path doesn't exists. See below for checking the path exists.

Only convert the string into a list here.

plexapi/library.py
```diff
@@ -294,8 +294,13 @@ def add(self, name='', type='', agent='', scanner='', location='', language='en'
...
+                raise BadRequest('Path: %s does not exist.' % location)
+            location = [location]
+        locations = [('location', l) for l in location]
```
JonnyWong16, line 301: Check the path exists for all locations in the list.

```py
locations = []
for path in location:
    try:
        self._server.browse(path)
    except NotFound:  # I'm not sure if this is the correct exception from `browse()`
        raise BadRequest('Path "%s" is not browsable by the Plex server')
    locations.append(('location', path))
```

plexapi/library.py
```diff
@@ -294,8 +294,13 @@ def add(self, name='', type='', agent='', scanner='', location='', language='en'
                   40:South Africa, 41:Spain, 42:Sweden, 43:Switzerland, 44:Taiwan, 45:Trinidad,
                   46:United Kingdom, 47:United States, 48:Uruguay, 49:Venezuela.
         """
-        part = '/library/sections?name=%s&type=%s&agent=%s&scanner=%s&language=%s&location=%s' % (
-            quote_plus(name), type, agent, quote_plus(scanner), language, quote_plus(location))  # noqa E126
+        if isinstance(location, str):
```
JonnyWong16, line 297: Update doc string to accept a list of locations.

plexapi/library.py
```diff
@@ -436,21 +441,80 @@ def reload(self, key=None):
...
+            Parameters:
+                location (str, list, or class): A single folder path, list of paths,
+                    or :class:`~plexapi.library.Location`.
+
+            Example:
+
+                .. code-block:: python
+                LibrarySection.addLocations('/path/1')
+                LibrarySection.addLocations(['/path/1', 'path/2', '/path/3'])
+                MovieLibrarySection.addLocations(ShowLibrarySection._locations()[0])
```
JonnyWong16, line 485: Avoid private methods in docs.

`MovieLibrarySection.addLocations(ShowLibrarySection.locations[0])`

Same for `removeLocations()`.

plexapi/library.py
```diff
@@ -436,21 +441,80 @@ def reload(self, key=None):
...
         # Reload this way since the self.key dont have a full path, but is simply a id.
         for s in self._server.library.sections():
             if s.key == self.key:
                 return s
 
+    def addLocations(self, location):
+        """ Add a location to a library.
+
+            Parameters:
+                location (str, list, or class): A single folder path, list of paths,
```
JonnyWong16, line 477: `location (str, list, or :class:`~plexapi.library.Location`)`

Same for `removeLocations()`.

plexapi/library.py
```diff
@@ -436,21 +441,80 @@ def reload(self, key=None):
...
+                .. code-block:: python
+                LibrarySection.addLocations('/path/1')
+                LibrarySection.addLocations(['/path/1', 'path/2', '/path/3'])
+                MovieLibrarySection.addLocations(ShowLibrarySection._locations()[0])
+        """
+        locations = self.locations
+        if isinstance(location, Location):
+            location = location.path
+        if isinstance(location, list):
+            locations += location
```
JonnyWong16, line 491: Use `locations.extend(location)`.

plexapi/library.py
```diff
@@ -436,21 +441,80 @@ def reload(self, key=None):
...
+                .. code-block:: python
+                LibrarySection.removeLocations('/path/1')
+                LibrarySection.removeLocations(['/path/1', 'path/2', '/path/3'])
+                LibrarySection.removeLocations(LibrarySection._locations()[0])
+        """
+        locations = self.locations
+        if isinstance(location, Location):
+            location = location.path
+        if isinstance(location, list):
+            locations -= location
```
JonnyWong16, line 513: Lists don't have a `-` method. You need to remove them using a for loop.

Also need to catch the `ValueError` and raise `BadRequest` when trying to remove a path that doesn't exist.

plexapi/library.py
```diff
@@ -436,21 +441,80 @@ def reload(self, key=None):
...
+        """ Edit a library. See :class:`~plexapi.library.Library` for example usage.
 
             Parameters:
+                agent (str, optional): The library agent.
                 kwargs (dict): Dict of settings to edit.
+
+            Example:
+
+                .. code-block:: python
+                LibrarySection.edit(**{'location': ['/path/1', 'path/2', '/path/3']})
```
JonnyWong16, line 453: Remove the example for editing locations. The convenience methods should be used since they check the path exists.

plexapi/library.py
```diff
@@ -436,21 +441,80 @@ def reload(self, key=None):
...
         """
         if not agent:
             agent = self.agent
-        part = '/library/sections/%s?agent=%s&%s' % (self.key, agent, urlencode(kwargs))
+
+        if kwargs['location']:
```
JonnyWong16, line 458:
```py
if 'location' in kwargs:
    # Assume paths have already been validated in addLocations and removeLocations
    locations = [('location', path) for path in kwargs.pop('location')]

params = list(kwargs.items()) + locations
```

plexapi/library.py
```diff
@@ -436,21 +441,80 @@ def reload(self, key=None):
...
+                MovieLibrarySection.addLocations(ShowLibrarySection._locations()[0])
+        """
+        locations = self.locations
+        if isinstance(location, Location):
+            location = location.path
+        if isinstance(location, list):
+            locations += location
+        else:
+            locations.append(location)
+        self.edit(location=locations)
```
JonnyWong16, line 494: Validate paths here using `self._server.browse(path)` before passing them to `edit()`.

JonnyWong16 (changes requested):

plexapi/library.py
blacktwin: I tried `.browse()` initially but iirc it returns `None` if the path is empty or does not exist.

plexapi/library.py
blacktwin: My example is showing how to use the `Location` class. In your example `ShowLibrarySection.locations[0]` returns the path as `str`. 

It could be that we don't need `Location` classes to used.

JonnyWong16: `Location` objects are not exposed to the user anywhere so it's unlikely they would come across it. If they do come across it, the class is listed as an accepted parameter in the docstring. Most users would use a location string since that is more natural.

plexapi/library.py
JonnyWong16: Hmm, I need to think about this.

plexapi/library.py
blacktwin: This assumes that a user would use the convenience method over `edit` for changing locations. Should the validation happen here instead then?

blacktwin:
Still needs tests.
- [ ] Create library with multiple paths with an invalid path
- [ ] Create library with multiple paths with valid paths
- [ ] Add path(s) to test library with valid paths
- [ ] Add path(s) to test library with an invalid path
- [ ] Remove path(s) to test library with valid paths
- [ ] Remove path(s) to test library with an invalid path

plexapi/library.py
```diff
@@ -294,8 +294,17 @@ def add(self, name='', type='', agent='', scanner='', location='', language='en'
                   40:South Africa, 41:Spain, 42:Sweden, 43:Switzerland, 44:Taiwan, 45:Trinidad,
                   46:United Kingdom, 47:United States, 48:Uruguay, 49:Venezuela.
         """
-        part = '/library/sections?name=%s&type=%s&agent=%s&scanner=%s&language=%s&location=%s' % (
-            quote_plus(name), type, agent, quote_plus(scanner), language, quote_plus(location))  # noqa E126
+        if isinstance(location, str):
+            if not self._server.isBrowsable(location):
```
JonnyWong16, line 298: Don't need browsable check here. Just convert to a list.

There's also a bug here if you don't specify any location, the default is `''`. I don't know what happens with `isBrowsable('')`.

plexapi/library.py
```diff
@@ -294,8 +294,17 @@ def add(self, name='', type='', agent='', scanner='', location='', language='en'
...
+                raise BadRequest('Path: %s does not exist.' % location)
+            location = [location]
+        if isinstance(location, list):
```
JonnyWong16, line 301: Don't need list check. Just start the for loop.

plexapi/library.py
```diff
@@ -294,8 +294,17 @@ def add(self, name='', type='', agent='', scanner='', location='', language='en'
...
+            for path in location:
+                if not self._server.isBrowsable(path):
+                    raise BadRequest('Path: %s does not exist.' % path)
+        locations = [('location', l) for l in location]
```
JonnyWong16, line 305: Create new locations list when checking browsable so you only loop once.

```suggestion
         _locations = []
        for path in location:
            if not self._server.isBrowsable(path):
                raise BadRequest('Path: %s does not exist.' % path)
            _locations.append(('location', path))
                
        part = '/library/sections?name=%s&type=%s&agent=%s&scanner=%s&language=%s&%s' % (
            quote_plus(name), type, agent, quote_plus(scanner), language, urlencode(_locations, doseq=True))
```

plexapi/library.py
```diff
@@ -436,21 +445,83 @@ def reload(self, key=None):
...
         if not agent:
             agent = self.agent
-        part = '/library/sections/%s?agent=%s&%s' % (self.key, agent, urlencode(kwargs))
+
+        locations = []
+        if kwargs['location']:
+            for path in kwargs['location']:
+                if not self._server.isBrowsable(path):
+                    raise BadRequest('Path: %s does not exist.' % path)
+            locations = [('location', path) for path in kwargs.pop('location')]
```
JonnyWong16, line 462: Append locations while checking browsable so you only loop once.

```suggestion
                locations.append(('location', path))
```

plexapi/library.py
```diff
@@ -436,21 +445,83 @@ def reload(self, key=None):
...
+        locations = self.locations
+        if isinstance(location, Path):
+            location = location.path
+        if isinstance(location, list):
+            locations = [path for path in locations if path not in location]
+        else:
+            locations.remove(location)
+        for path in locations:
+            if not self._server.isBrowsable(path):
+                raise BadRequest('Path: %s does not exist.' % path)
```
JonnyWong16, line 522: The exception should be something like "The path you are trying to remove does not exist in the library". Presumably the current paths in the library are already browsable and don't need to be checked when removing.

JonnyWong16 (changes requested):

plexapi/library.py
blacktwin: Removed the extra browseable check for the string. Looks like it's covered to me. 
```python
plex.library.add(name='Test Library', type='show', agent='tv.plex.agents.series',
                     scanner='Plex TV Series', location='', language='en-US')
# plexapi.exceptions.BadRequest: Path:  does not exist.
```

plexapi/library.py
blacktwin: We need to check for locations that are str in order to loop properly.
```suggestion
        locations = []
        if kwargs['location']:
            if isinstance(kwargs['location'], str):
                kwargs['location'] = [kwargs['location']]
            for path in kwargs.pop('location'):
                if not self._server.isBrowsable(path):
                    raise BadRequest('Path: %s does not exist.' % path)
                locations.append(('location', path))
```

plexapi/library.py
blacktwin: d1b1903 and 5cf95e5

plexapi/library.py
```diff
@@ -486,16 +494,88 @@ def reload(self):
...
+                if not self._server.isBrowsable(path):
+                    raise BadRequest('Path: %s does not exist.' % path)
+                locations.append(('location', path))
+
+        params = list(kwargs.items()) + locations
+
+        part = '/library/sections/%s?agent=%s&%s' % (self.key, agent, urlencode(params, doseq=True))
         self._server.query(part, method=self._server._session.put)
 
+        # Reload this way since the self.key dont have a full path, but is simply a id.
```
JonnyWong16, line 520: This reloading was removed in #855.

plexapi/library.py
```diff
@@ -486,16 +494,88 @@ def reload(self):
...
+                LibrarySection.removeLocations('/path/1')
+                LibrarySection.removeLocations(['/path/1', 'path/2', '/path/3'])
+                LibrarySection.removeLocations(PlexServer.browse()[0])
+        """
+        locations = self.locations
+        if isinstance(location, Path):
+            location = location.path
+        if isinstance(location, str):
+            location = [location]
+        for _location in location:
```
JonnyWong16, line 570:
```suggestion
        for path in location:
```

JonnyWong16 (changes requested):
Few more tests:
- [x] `BadRequest` when trying to remove all locations.
- [ ] `edit` directly with `kwargs['location']`.
- [x] `addLocations` with a `Path` object.
- [ ] `removeLocations` with a `Path` object.
- [ ] `isBrowsable` with a `Path` object.

blacktwin:
I don't think we should include `Path` object for adding or removing locations. `plex.browse('path/to/stuff')` returns a list of the `Path` objects from in the queried path but not the `Path` object for the queried path. This will either fail if the queried path does not have any subfolder or would result in adding each subfolder instead of the queried path when using `.addLocation(Path)`.

plexapi/library.py
```diff
@@ -169,7 +169,7 @@ def add(self, name='', type='', agent='', scanner='', location='', language='en'
                 name (str): Name of the library
                 agent (str): Example com.plexapp.agents.imdb
                 type (str): movie, show, # check me
-                location (str): /path/to/files
+                location (str or list): /path/to/files, [/path/to/files, /path/to/morefiles]
```
JonnyWong16, line 172:
```suggestion
                location (str or list): /path/to/files, ["/path/to/files", "/path/to/morefiles"]
```

tests/test_library.py
```diff
@@ -121,24 +121,63 @@ def test_library_recentlyAdded(plex):
...
+        section.addLocations(movie_location[:-1])
+    with pytest.raises(BadRequest):
+        section.removeLocations(movie_location[:-1])
+    section.removeLocations(photo_location)
+    section.reload()
+    assert len(section.locations) == 1
+    section.addLocations(photo_location)
+    section.reload()
+    assert len(section.locations) == 2
+    section.edit(**{'location': [movie_location]})
```
JonnyWong16, line 175:
```suggestion
    section.edit(**{'location': movie_location})
```

Add one more test for `edit` with a `BadRequest` for bad location.

plexapi/library.py
```diff
@@ -486,16 +494,75 @@ def reload(self):
...
+                raise BadRequest('Path: %s does not exist.' % path)
+            locations.append(path)
+        self.edit(location=locations)
+
+    def removeLocations(self, location):
+        """ Remove a location from a library.
+        
+            Parameters:
+                location (str or list): A single folder path, list of paths.
+            Example:
```
JonnyWong16, line 547: Blank line before example.

JonnyWong16 (changes requested):

JonnyWong16 (approved):

PR #870, opened 2022-01-12
Support for `Artist` radio stations

jjlawren:
## Description

This adds support for artist radio stations. In Plex Web, this is represented by the "Play Radio" button on an artist page.

An artist radio station can be retrieved using `Artist.station()`, which returns a `Playlist` if it exists. To differentiate these radio stations from classic playlists, a new `radio` boolean attribute is exposed.

`PlayQueue` creation has been updated to handle these special radio `Playlist` instances, which allows `PlexClient.playMedia` to accept them as-is.

## Type of change

## Checklist:

### DIFF
--- a/plexapi/playlist.py
+++ b/plexapi/playlist.py
@@ -29,7 +29,11 @@ class Playlist(PlexPartialObject, Playable, ArtMixin, PosterMixin, SmartFilterMi
             icon (str): Icon URI string for smart playlists.
             key (str): API URL (/playlist/<ratingkey>).
             leafCount (int): Number of items in the playlist view.
+            librarySectionID (int): Library section identifier (radio only)
+            librarySectionKey (str): Library section key (radio only)
+            librarySectionTitle (str): Library section title (radio only)
             playlistType (str): 'audio', 'video', or 'photo'
+            radio (bool): If this playlist represents a radio station
             ratingKey (int): Unique key identifying the playlist.
             smart (bool): True if the playlist is a smart playlist.
             summary (str): Summary of the playlist.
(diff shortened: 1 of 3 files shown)


plexapi/playlist.py
```diff
@@ -54,7 +58,10 @@ def _loadData(self, data):
         self.icon = data.attrib.get('icon')
         self.key = data.attrib.get('key', '').replace('/items', '')  # FIX_BUG_50
         self.leafCount = utils.cast(int, data.attrib.get('leafCount'))
+        self.librarySectionKey = data.attrib.get('librarySectionKey')
```
JonnyWong16, line 61: Added `librarySectionID` to doc string but missing attribute?
```suggestion
        self.librarySectionID = utils.cast(int, data.attrib.get('librarySectionID'))
        self.librarySectionKey = data.attrib.get('librarySectionKey')
```

plexapi/playqueue.py
```diff
@@ -175,8 +175,11 @@ def create(
             args["uri"] = "library:///directory/{uri_args}".format(uri_args=uri_args)
             args["type"] = items[0].listType
         elif items.type == "playlist":
-            args["playlistID"] = items.ratingKey
             args["type"] = items.playlistType
+            if items.radio:
+                args["uri"] = f"server://{server.machineIdentifier}/{server.library.identifier}{items.key}"
```
JonnyWong16, line 180: Use `%s` string formatting to be consistent with the rest of Plex API?
```suggestion
                args["uri"] = "server://%s/%s%s" % (
                    server.machineIdentifier, server.library.identifier, items.key
                )
```

JonnyWong16 (changes requested):

plexapi/playqueue.py
JonnyWong16: I missed all the other strings in `playqueue` all use `.format()`.

jjlawren: I'd prefer to use f-strings moving forward. Most string formatting in the project was written before we could use the feature in all supported Python versions. Now that 3.5 support has dropped they can be used safely.

JonnyWong16: I would too, but I want to do it all in one go.

jjlawren: It's a style change that doesn't affect functionality. Let's require f-strings for any additions/changes moving forward and convert existing formatting in followup PRs when time/effort permits.

plexapi/playlist.py
jjlawren: This is already set elsewhere, but it _is_ included in the playlist XML payload. Better to set it explicitly like you suggest. 👍

JonnyWong16 (approved):

PR #868, opened 2022-01-10
Feature: Add label methods, properties to Episodes

glensc:
## Description

Seems `LabelMixin` was added to `Album`, `Collection`, `Show`, and `Movie`, but left out of `Episode`.

My testing with `1.24.3.5033-757abe6b4` shows that Episodes can also have labels.

```py
>>> episode.addLabel(['skip-pts'])
>>> episode = (fetch episode again)
>>> print(episode.labels)
[<Label:/library/sections/2/:Skip-pts>]
```

## Type of change

Please delete options that are not relevant.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## Checklist:

- [ ] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have added or updated the docstring for new or existing methods
- [x] I have added tests when applicable

PR #872, opened 2022-01-13
Add LabelMixin to Season, Episode, Artist, and Track objects

meisnate12:
## Description

Following up on #868 this adds the `LabelMixin` to the `Season`, `Episode`, `Artist`, and `Track` objects as well. 

## Type of change

## Checklist:

### DIFF
--- a/plexapi/audio.py
+++ b/plexapi/audio.py
@@ -125,7 +125,7 @@ class Audio(PlexPartialObject):
 
 @utils.registerPlexObject
 class Artist(Audio, AdvancedSettingsMixin, ArtMixin, PosterMixin, RatingMixin, SplitMergeMixin, UnmatchMatchMixin,
-        CollectionMixin, CountryMixin, GenreMixin, MoodMixin, SimilarArtistMixin, StyleMixin):
+        CollectionMixin, CountryMixin, GenreMixin, LabelMixin, MoodMixin, SimilarArtistMixin, StyleMixin):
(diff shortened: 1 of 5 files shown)


JonnyWong16:
Has anyone checked that applying these hidden labels doesn't break sharing restrictions for shared users?

meisnate12:
I didn't test music libraries but for shows I add a label to a season and an episode and restricted that label from a test user and the user was still able to view both the season and the episode, which leads me to believe they're just ignored. and when the label is added to the show the whole thing is ignored

JonnyWong16:
My other concern is users can get stuck with labels that they have no idea how to remove.

Just as an example, if Plex-Meta-Manager adds a feature that can label episodes they are now stuck with the labels unless they know how to remove them again with Plex API. Those hidden labels will continue to show up in search filters and sharing restrictions. There is no way to remove those labels in the UI.

It's part of the same reason why #510 and #658 have not been merged.

meisnate12:
I mean thats a very valid concern as my plan was to add labels to seasons and episodes in order to be able to tell what has had an image overlay on it and what hasn't. Ill include a way to remove all labels using PMM so if they got into the situation with PMM they could at least get back out using the same software.

JonnyWong16:
Please test my changes make labels a searchable filter.

meisnate12:
i ran a few tests on each type of item and they were all searchable using filters

Issue #671, opened 2021-02-20
plexapi tidal integration


62274677:
Does the plexapi include/are there plans to include tidal support? I am looking to write a script that syncs a playqueue between youtube and plex/plexamp. If I can only find the currently playing songs and up next from plex, that is alright, but it would be wonderful if I could also add tidal songs to the playqueue programmatically. I am new to the plexapi, so if this exists already I apologize; I couldn't find any reference to tidal other than a single command in the docs.

darovic:
I was going to try this, but my head is swimming looking through the Plex logs trying to figure out the relevant APIs.

Does the attached log file assist in any way? Amongst allt he noise, it has me:
1. creating a new playlist named `SomePlaylistName`
2. adding a Tidal song to my Library (`Jericho` by `Celldweller`)
3. adding that song to the playlist `SomePlaylistName`
4. playing that song

[partial_log.txt](https://github.com/pkkid/python-plexapi/files/8077259/partial_log.txt)

jjlawren:
@darovic have you tried with the extra header suggestion from #905?

trankillity:
Just adding my voice that I'd love to see this integrated. Even if it's just to a basic level with track details coming through, not necessarily any library API support. Using the Home Assistant Plex integration (which I believe uses `python-plexapi`) just shows the last local track played in the entity attributes.

PR #905, opened 2022-03-05
Enable external media in responses, e.g. Tidal

zacwest:
## Description

Adds the header `X-Plex-Features` which controls whether Tidal results are included in responses.

For example, running against my Plex server:

```python
# 61115 for track.source and album.source is for items from Tidal
len(music.searchAlbums(filters = {'album.source': '61115'}))
```

Results in `0` before this change and non-`0` after.

Refs #671.

The value that I see in the web frontend is `external-media,indirect-media` but the latter does not appear to change the results that I can tell; erring on the side of not adding it since it may cause another level of effects.

I don't know the Plex API well enough to know whether this may have additional side effects that I'm not considering. If that's the case, it may be worth making this opt-in instead.

## Type of change

## Checklist:

### DIFF
--- a/plexapi/config.py
+++ b/plexapi/config.py
@@ -62,4 +62,5 @@ def reset_base_headers():
         'X-Plex-Device-Name': plexapi.X_PLEX_DEVICE_NAME,
         'X-Plex-Client-Identifier': plexapi.X_PLEX_IDENTIFIER,
         'X-Plex-Sync-Version': '2',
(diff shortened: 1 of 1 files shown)


jjlawren (approved):
Thanks for tracking this down 👍 

I haven't run into any obvious negative side-effects during testing, but I don't have an active Tidal subscription. If issues are reported this can always be moved to an opt-in setting.

PR #1000, opened 2022-08-01
Update PlexAPI to f-strings

JonnyWong16:
## Description

Updates all string formatting to use f-strings.

Used `flynt` to convert f-strings.
https://github.com/ikamensh/flynt

## Type of change

## Checklist:

### DIFF
--- a/plexapi/audio.py
+++ b/plexapi/audio.py
@@ -120,7 +120,7 @@ class Audio(PlexPartialObject):
 
         section = self._server.library.sectionByID(self.librarySectionID)
 
-        sync_item.location = 'library://%s/item/%s' % (section.uuid, quote_plus(self.key))
+        sync_item.location = f'library://{section.uuid}/item/{quote_plus(self.key)}'
         sync_item.policy = Policy.create(limit)
         sync_item.mediaSettings = MediaSettings.createMusic(bitrate)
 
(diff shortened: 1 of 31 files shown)


plexapi/audio.py
```diff
@@ -120,7 +120,7 @@ def sync(self, bitrate, client=None, clientId=None, limit=None, title=None):
 
         section = self._server.library.sectionByID(self.librarySectionID)
 
-        sync_item.location = 'library://%s/item/%s' % (section.uuid, quote_plus(self.key))
+        sync_item.location = f'library://{section.uuid}/item/{quote_plus(self.key)}'
```
Hellowlol, line 123: Isnt the curlybrace at the wrong place? Shouldnt it be quoute_plus({self.key})

Hellowlol: F strings should not be used in logging for performance reasons.

JonnyWong16: > Isnt the curlybrace at the wrong place? Shouldnt it be quoute_plus({self.key})

No, the brackets are correct. It has to evaluate `quote_plus` before formatting.

> F strings should not be used in logging for performance reasons.

The performance difference (faster or slower) is negligible.

https://github.com/PyCQA/pylint/issues/2354#issuecomment-414526879

PR #1039, opened 2022-10-03
Add timeout parameter to `uploadTheme`

ReenigneArcher:
## Description

Please include a summary of the change and which issue is fixed.

Since themes may take a while to upload, it may be beneficial to have a separate `timeout` option for uploading themes, instead of using the default `config.TIMEOUT`.

## Type of change

## Checklist:


JonnyWong16:
I'm not too keen on adding a one-off `timeout` parameter. There must be a better solution to this.

ReenigneArcher:
I'm not sure if there is a better solution. The options I know of are to set a global timeout, or set it in the query. I don't think it makes sense to set it globally to something that is very high. For my use case I now have it set to 5 minutes to handle theme uploading. Setting it to 5 minutes globally would be annoying if you have to wait that long to find out that the server may be offline.

One other idea... Every class that has a `server.query` method could perhaps accept kwargs instead so that the programmer can select whatever query parameters they prefer.

ReenigneArcher:
I probably should have given more context as to why this is an issue to begin with.

Currently, if you hit the timeout while uploading a theme you get a partially uploaded theme. It's still uploaded and will play within the Plex UI, but it will abruptly cut off at the point where the timeout occurred.

I doubt the 30s second timeout is an issue with any of the other uploading mixins, since photos are much smaller.

Issue #1067, opened 2022-12-18
markPlayed and markUnplayed Error for medias from Plex Discover


twolaw:
### Describe the Bug

Plex Discover allows to get Movies and Shows that user don't have in PMS library.
Problem : the `markPlayed()` and `markUnplayed()` functions do not work for those medias found with `discover.provider.plex.tv` it raises a __plexapi.exceptions.NotFound__.

Movies and Shows (found from Plex Discover) watched status can be switched with Plex Client, we shoud be able to do it through Plex API too.

### Code Snippets

```python
>>> myMovie
<Movie:nan:Cairo-Conspiracy>
>>> myMovie.markPlayed()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "C:\Users\me\AppData\Local\Packages\PythonSoftwareFoundation.Python.3.9_qbz5n2kfra8p0\LocalCache\local-packages\Python39\site-packages\plexapi\mixins.py", line 274, in markPlayed
    self._server.query(key, params=params)
  File "C:\Users\me\AppData\Local\Packages\PythonSoftwareFoundation.Python.3.9_qbz5n2kfra8p0\LocalCache\local-packages\Python39\site-packages\plexapi\server.py", line 705, in query
    raise NotFound(message)
plexapi.exceptions.NotFound: (404) not_found; https://metadata.provider.plex.tv/:/scrobble?key=nan&identifier=com.plexapp.plugins.library <Error error="Not Found" message="Not Found" statusCode="404"/>
```

### Expected Behavior

No error raised.
Just like items from user's PMS library.

### Additional Context

Requests to switch watched state of items from Plex Discover are :
```
https://discover.provider.plex.tv/actions/scrobble
https://discover.provider.plex.tv/actions/unscrobble
```

with data:
```
key=(ratingKey)
X-Plex-Token=(accessToken)
...
```

And we can check viewCount and lastViewdAt with request :
`https://metadata.provider.plex.tv/library/metadata/(ratingKey)/userState`

I think this uses the same Plex cloud database (linked to user account) to [sync PMS watched state and ratings](https://support.plex.tv/articles/sync-watch-state-and-ratings/).

### Operating System and Version

Windows 10

### Plex Media Server Version

1.29.2.6364

### Python Version

3.7

### PlexAPI Version

4.13.1

RileyXX:
Using Plex Discover to mark items as played/unplayed is a clever way to do this. +1 for this issue

[...]

PR #1090, opened 2023-01-11
Add support Discover isPlayed/markPlayed/markUnplayed

glensc:
## Description

Fixes https://github.com/pkkid/python-plexapi/issues/1067

## Type of change

## Checklist:

### DIFF
--- a/plexapi/myplex.py
+++ b/plexapi/myplex.py
@@ -937,7 +937,7 @@ class MyPlexAccount(PlexObject):
         return self.findItem(data, cls=UserState)
 
     def isPlayed(self, item):
-        """ Returns True if this item is played.
+        """ Return True if the item is played on Discover.
 
             Parameters:
                 item (:class:`~plexapi.video.Movie`,
@@ -949,7 +949,7 @@ class MyPlexAccount(PlexObject):
         return bool(userState.viewCount > 0) if userState.viewCount else False
 
     def markPlayed(self, item):
-        """ Mark the Plex object as played.
+        """ Mark the Plex object as played on Discover.
 
             Parameters:
                 item (:class:`~plexapi.video.Movie`,
@@ -964,7 +964,7 @@ class MyPlexAccount(PlexObject):
         return self
 
     def markUnplayed(self, item):
-        """ Mark the Plex object as unplayed.
+        """ Mark the Plex object as unplayed on Discover.
 
             Parameters:
                 item (:class:`~plexapi.video.Movie`,


glensc:
need to add test to see if isWatched needs to be hacked too to fetch from myPlex

glensc:
> need to add test to see if isWatched needs to be hacked too to fetch from myPlex. or rather throw to suggest to use myPlex directly for online media?

yes. media.isPlayed does not have online source metadata, so after marking played in cloud, the property is always False.

here's my test script

```python
from plexapi.video import Movie


def test_discover(env):
    url = env.get('PLEX_URL')
    token = env.get('PLEX_TOKEN')
    plex = PlexServer(baseurl=url, token=token)
    # https://app.plex.tv/desktop/#!/provider/tv.plex.provider.discover/details?key=%2Flibrary%2Fmetadata%2F5d77696a7a53e9001e6e5da0
    result = account.searchDiscover("Olympus Has Fallen", libtype="movie")
    movie: Movie = result[0]

    print(movie)
    print("isPlayed: ", movie.isPlayed)
    print(movie.markPlayed())
    print("isPlayed: ", movie.isPlayed)


if __name__ == "__main__":
    import os
    import logging

    from plexapi import log as logger
    from plexapi import loghandler

    logger.removeHandler(loghandler)
    logger.setLevel(logging.DEBUG)
    logging.basicConfig(format="%(asctime)s:%(name)s: %(message)s", level=logging.DEBUG)

    test_discover(os.environ)

```

glensc:
this is an ugly hack, I'm sure this should be solved differently, perhaps by adding a new Movie type, DiscoverMovie that has similar properties as Movie.

Or just leave as is and do the code in user code (outside PlexAPI)

glensc:
here's one thing I noticed when testing, `userState.viewCount` gets incremented every time `markPlayed()` is called, but `markUnplayed()` resets play count to 0, not just decrementing

glensc:
or should the API be similar to `MyPlexAccount.userState`?

```py
server = plex.myPlexAccount()
server.markPlayed(movie)
server.markUnplayed(movie)
```

glensc:
ok, at 035c446, can use myplex. this seems more cleaner approach than hacking `PlayedUnplayedMixin`

```python
from plexapi.video import Movie


def test_discover(env):
    url = env.get('PLEX_URL')
    token = env.get('PLEX_TOKEN')
    plex = PlexServer(baseurl=url, token=token)
    # https://app.plex.tv/desktop/#!/provider/tv.plex.provider.discover/details?key=%2Flibrary%2Fmetadata%2F5d77696a7a53e9001e6e5da0
    result = account.searchDiscover("Olympus Has Fallen", libtype="movie")
    movie: Movie = result[0]

    print(movie)
    print(movie.guid)
    print("isPlayed: ", account.isPlayed(movie))
    print(account.markPlayed(movie))
    print("isPlayed: ", account.isPlayed(movie))
    print(account.markUnplayed(movie))
    print("isPlayed: ", account.isPlayed(movie))


if __name__ == "__main__":
    import os
    import logging

    from plexapi import log as logger
    from plexapi import loghandler

    logger.removeHandler(loghandler)
    logger.setLevel(logging.DEBUG)
    logging.basicConfig(format="%(asctime)s:%(name)s: %(message)s", level=logging.DEBUG)

    test_discover(os.environ)
```

glensc:
Maintainers: What version do I keep then? myPlex or PlayedUnplayedMixin?

I think myPlex is cleaner:
- https://github.com/pkkid/python-plexapi/pull/1090#issuecomment-1378971206

JonnyWong16:
I agree. Use the myplex account as a temporary solution until proper support for Discover can be added.

glensc:
Cleaned up

plexapi/myplex.py
```diff
@@ -936,6 +936,27 @@ def userState(self, item):
         data = self.query(f"{self.METADATA}/library/metadata/{ratingKey}/userState")
         return self.findItem(data, cls=UserState)
 
+    def isPlayed(self, item):
+        """ Returns True if this item is played. """
```
JonnyWong16, line 940: Add documentation for the `item` parameter.

Explicitly state that `item` must be a video object from `searchDiscover()`.

Same for the other two methods.

JonnyWong16 (changes requested):

plexapi/myplex.py
glensc: item can be also video object if it's from new plex agent. i.e media object with guid `plex://<hash>`

JonnyWong16: Right, I forgot about that. Then just the same as the `item` parameter from `userState()` above.

But update the doc sting to say "Return True if the item is played on Discover." (same for the other two methods) so it doesn't get confused with the mixin methods.

glensc: Added: e986b56

agents doc:
- https://support.plex.tv/articles/200241558-agents/

glensc: Added discover note: c61c733

JonnyWong16 (approved):

RileyXX:
Great work guys! Thanks for the implementation.

PR #1155, opened 2023-05-08
Read version from file instead of import in setup.py

JonnyWong16:
## Description

Read version from file instead of import in `setup.py`.

User packages (i.e. `requests`) are no longer available during the build process which causes a `ModuleNotFound` error.

Ref: https://packaging.python.org/en/latest/guides/single-sourcing-package-version/

## Type of change

## Checklist:

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -9,10 +9,14 @@ try:
 except ImportError:
     from distutils.core import setup
 
-from plexapi import const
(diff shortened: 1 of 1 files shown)


JonnyWong16:
We should migrate to a `pyproject.toml` file at some point.

PR #1209, opened 2023-08-06
Fix ElementTree DeprecationWarning when testing truth value (3.12)

cdce8p:
## Description
Fix a DeprecationWarning for Python 3.12.
```
  /.../python3.12/site-packages/plexapi/client.py:97: DeprecationWarning: Testing an element's truth value will raise an exception in future versions.  Use specific 'len(elem)' or 'elem is not None' test instead.
    if not data:
```

From the Python 3.12 [Changelog](https://docs.python.org/3.12/whatsnew/3.12.html#deprecated)

> [xml.etree.ElementTree](https://docs.python.org/3.12/library/xml.etree.elementtree.html#module-xml.etree.ElementTree): The module now emits [DeprecationWarning](https://docs.python.org/3.12/library/exceptions.html#DeprecationWarning) when testing the truth value of an [xml.etree.ElementTree.Element](https://docs.python.org/3.12/library/xml.etree.elementtree.html#xml.etree.ElementTree.Element). Before, the Python implementation emitted [FutureWarning](https://docs.python.org/3.12/library/exceptions.html#FutureWarning), and the C implementation emitted nothing.

https://github.com/python/cpython/blob/v3.12.0rc1/Lib/xml/etree/ElementTree.py#L201-L208

## Type of change

### DIFF
--- a/plexapi/client.py
+++ b/plexapi/client.py
@@ -94,7 +94,7 @@ class PlexClient(PlexObject):
             raise Unsupported('Cannot reload an object not built from a URL.')
         self._initpath = self.key
         data = self.query(self.key, timeout=timeout)
-        if not len(data):
+        if data is None:
             raise NotFound(f"Client not found at {self._baseurl}")
         if self._clientIdentifier:
             client = next(
--- a/plexapi/client.py
+++ b/plexapi/client.py
@@ -94,7 +94,7 @@ class PlexClient(PlexObject):
             raise Unsupported('Cannot reload an object not built from a URL.')
         self._initpath = self.key
         data = self.query(self.key, timeout=timeout)
-        if not data:
+        if not len(data):
             raise NotFound(f"Client not found at {self._baseurl}")
         if self._clientIdentifier:
             client = next(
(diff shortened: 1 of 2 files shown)


jjlawren (approved):

jjlawren:
Thanks, @cdce8p!

plexapi/client.py
```diff
@@ -94,7 +94,7 @@ def connect(self, timeout=None):
             raise Unsupported('Cannot reload an object not built from a URL.')
         self._initpath = self.key
         data = self.query(self.key, timeout=timeout)
-        if not data:
+        if not len(data):
```
JonnyWong16, line 97: Should this be `if data is not None` because `len(None)` will raise `TypeError`?

JonnyWong16 (changes requested):

plexapi/client.py
cdce8p: Maybe `if data is None`? I assume this should match the error case.

JonnyWong16:
I found two more locations.

https://github.com/pkkid/python-plexapi/blob/a7d29a3638bdbec6e223c373a47aeb36a2e60bd5/plexapi/myplex.py#L1776

https://github.com/pkkid/python-plexapi/blob/a7d29a3638bdbec6e223c373a47aeb36a2e60bd5/plexapi/myplex.py#L1793

JonnyWong16 (approved):

cdce8p:
Anything left I should do here?

Issue #428, opened 2020-02-17
Fewer hard dependencies


coleifer:
First off, thank you very much for this *excellent* library.

I use plexapi for a number of scripts that I run as full-fledged applications on my system. This means that I need to install plexapi's dependencies where my system python can find them. There are two dependencies which I think could be made optional: `tqdm` and the `websocket` client.

`tqdm` is only needed if you use the download feature *and* show status. I've patched this locally as a proof of how easy it is to drop tqdm as a hard dependency.

The `websocket` client is only needed if you intend to subscribe to alerts. It would be nice if the absence of this library could be handled gracefully, as I have no interest in the alerts features and don't want to install this dep.

Here's a simple patch I'm using locally:

```diff
diff --git a/plexapi/alert.py b/plexapi/alert.py
index 38dc930..be78e7c 100644
--- a/plexapi/alert.py
+++ b/plexapi/alert.py
@@ -1,7 +1,10 @@
 # -*- coding: utf-8 -*-
 import json
 import threading
-import websocket
+try:
+    import websocket
+except ImportError:
+    websocket = None
 from plexapi import log
 
 
diff --git a/plexapi/utils.py b/plexapi/utils.py
index e8ff989..5b0a1a7 100644
--- a/plexapi/utils.py
+++ b/plexapi/utils.py
@@ -8,7 +8,10 @@ import zipfile
 from datetime import datetime
 from getpass import getpass
 from threading import Thread, Event
-from tqdm import tqdm
+try:
+    from tqdm import tqdm
+except ImportError:
+    tqdm = None
 from plexapi import compat
 from plexapi.exceptions import NotFound
 
@@ -287,6 +290,9 @@ def download(url, token, filename=None, savepath=None, session=None, chunksize=4
 
     # save the file to disk
     log.info('Downloading: %s', fullpath)
+    if tqdm is None:
+        showstatus = False
+
     if showstatus:  # pragma: no cover
         total = int(response.headers.get('content-length', 0))
         bar = tqdm(unit='B', unit_scale=True, total=total, desc=filename)
```

pkkid:
I think the only thing I would add to these patches is a log and print message is `websocket is None` when trying to subscribe to alerts.

coleifer:
If there's interest in such a patch, I'll be happy to submit a cleaned-up version. I just wanted to float the idea first.

Montellese:
[...]

PR #1219, opened 2023-08-19
feat(alert): add option to use custom socket

ReenigneArcher:
## Description

This PR adds the ability to pass in a custom socket for the AlertListener.

In some cases attempting to use the AlertListener, will fail with the message `getsockaddrarg: bad family`. I believe this is due to using IPv6 on Linux hosts, but I cannot confirm. Adding the ability to use a custom socket, would allow the developer to create their own socket to use in this event.

## Type of change

## Checklist:

### DIFF
--- a/plexapi/alert.py
+++ b/plexapi/alert.py
@@ -1,5 +1,7 @@
 # -*- coding: utf-8 -*-
 import json
+import socket
+from typing import Callable
 import threading
 
 from plexapi import log
@@ -32,15 +34,17 @@ class AlertListener(threading.Thread):
             callbackError (func): Callback function to call on errors. The callback function
                 will be sent a single argument 'error' which will contain the Error object.
                 :samp:`def my_callback(error): ...`
+            ws_socket (socket): Socket to use for the connection. If not specified, a new socket will be created.
     """
     key = '/:/websockets/notifications'
 
-    def __init__(self, server, callback=None, callbackError=None):
+    def __init__(self, server, callback: Callable = None, callbackError: Callable = None, ws_socket: socket = None):
         super(AlertListener, self).__init__()
         self.daemon = True
         self._server = server
         self._callback = callback
         self._callbackError = callbackError
+        self._socket = ws_socket
         self._ws = None
 
     def run(self):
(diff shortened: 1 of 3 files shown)


plexapi/alert.py
```diff
@@ -52,8 +56,13 @@ def run(self):
...
         url = self._server.url(self.key, includeToken=True).replace('http', 'ws')
         log.info('Starting AlertListener: %s', url)
-        self._ws = websocket.WebSocketApp(url, on_message=self._onMessage,
-                                          on_error=self._onError)
+
+        if self._socket:
+            self._ws = websocket.create_connection(url=url, on_message=self._onMessage, on_error=self._onError,
+                                                   socket=self._socket)
+        else:
+            self._ws = websocket.WebSocketApp(url, on_message=self._onMessage, on_error=self._onError)
```
JonnyWong16, line 64: `WebSocketApp` accepts a `socket` argument.

https://websocket-client.readthedocs.io/en/latest/app.html#websocket._app.WebSocketApp

ReenigneArcher: It doesn't in earlier versions. I assumed you wanted to keep compatibility with older versions based on this comment, below.

```py
    def _onMessage(self, *args):
        """ Called when websocket message is received.
            In earlier releases, websocket-client returned a tuple of two parameters: a websocket.app.WebSocketApp
            object and the message as a STR. Current releases appear to only return the message.
            We are assuming the last argument in the tuple is the message.
            This is to support compatibility with current and previous releases of websocket-client.
        """
```

I can update it if you prefer though?

JonnyWong16: `create_connection` returns a `WebSocket` object, so it's not quite the same as `WebSocketApp` object.

I think it's fine to target the latest version of `websocket-client`. They actually reverted back to returning a two-tuple in the latest version so the docstring for `_onMessage` and `_onError` is wrong anyways. I think the entire comment about previous versions of `websocket-client` can be removed.

ReenigneArcher: To avoid a bunch if if/else logic, I added `websocket-client>=1.3.3` to the `requirements.txt`. It was also kind of annoying to have to install it separately. I think there's a way to set optional dependencies with the setup.py, but I'm not sure how... or how to have pip install them.

requirements.txt
```diff
@@ -3,3 +3,4 @@
 # pip install -r requirements.txt
 #---------------------------------------------------------
 requests
+websocket-client>=1.3.3
```
JonnyWong16, line 6: `websocker-client` was removed from `requirements.txt` in #468 (and discussion in #428).

JonnyWong16 (changes requested):

JonnyWong16 (approved):

PR #1239, opened 2023-09-09
Update searchDiscover to use discover baseurl instead of metadata

twolaw:
## Description

To search content in Plex Discover service, we used to request `https://metadata.provider.plex.tv/library/search` but now the url has changed to `https://discover.provider.plex.tv/library/search`.
Old url always returns *404 Not found*.

`metadata.provider.plex.tv` hostname still works to fetch online metadata and watchlist but not used anymore by Plex Web app.

Fixes https://github.com/Taxel/PlexTraktSync/issues/1575

## Type of change

## Checklist:

### DIFF
--- a/plexapi/myplex.py
+++ b/plexapi/myplex.py
@@ -111,6 +111,7 @@ class MyPlexAccount(PlexObject):
     # Hub sections
     VOD = 'https://vod.provider.plex.tv'                                                        # get
     MUSIC = 'https://music.provider.plex.tv'                                                    # get
+    DISCOVER = 'https://discover.provider.plex.tv'
     METADATA = 'https://metadata.provider.plex.tv'
     key = 'https://plex.tv/api/v2/user'
 
@@ -1056,7 +1057,7 @@ class MyPlexAccount(PlexObject):
             'includeMetadata': 1
         }
 
-        data = self.query(f'{self.METADATA}/library/search', headers=headers, params=params)
+        data = self.query(f'{self.DISCOVER}/library/search', headers=headers, params=params)
         searchResults = data['MediaContainer'].get('SearchResults', [])
         searchResult = next((s.get('SearchResult', []) for s in searchResults if s.get('id') == 'external'), [])
 


ReenigneArcher:
I can confirm, claimed tests now have 8 failures. https://github.com/LizardByte/python-plexapi-backport/actions/runs/6132718396/job/16643914590#step:12:1964

JonnyWong16:
Should we pre-emptively change the rest of the methods to `discover.provider.plex.tv`? I just browsed through Plex Web and it seems like everything is switched over, including playlists and metadata.

twolaw:
I agree.
I keep METADATA const just in case this baseurl gets back later.

plexapi/myplex.py
```diff
@@ -1117,9 +1118,9 @@ def link(self, pin):
 
     def _toOnlineMetadata(self, objs, **kwargs):
         """ Convert a list of media objects to online metadata objects. """
-        # TODO: Add proper support for metadata.provider.plex.tv
+        # TODO: Add proper support for discover.provider.plex.tv
         # Temporary workaround to allow reloading and browsing of online media objects
-        server = PlexServer(self.METADATA, self._token, session=self._session)
+        server = PlexServer(self.DISCOVER, self._token, session=self._session)
```
JonnyWong16, line 1123: This might be the one exception to leave as `METADATA`.

twolaw: I don't understand why.
All Plex Web app requests are sent to `DISCOVER`.

JonnyWong16: And now that I think about it some more, you can add unrated content to your watchlist and also mark it played/unplayed. I don't know how the Plex UI would handle this, but it works through the API.

So maybe only the `searchDiscover` URL should change, and just leave the rest of them as `metadata`.

twolaw: Ok, back to first commit only.

twolaw:
Listed with a wrong PR number in 4.15.2 release changelog.

PR #1240, opened 2023-09-09
ci: retry bootstrap step

ReenigneArcher:
## Description
Fixes ci issue where bootstrapping server would randomly fail. It will now retry up to 3 times.

## Type of change

- [x] None of the original options

## Checklist:

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -157,15 +157,26 @@ jobs:
         echo "PLEXAPI_AUTH_SERVER_TOKEN=${{ secrets.PLEXAPI_AUTH_SERVER_TOKEN }}" >> $GITHUB_ENV
 
     - name: Bootstrap ${{ matrix.plex }} Plex server
-      run: |
-        . venv/bin/activate
-        python \
-          -u tools/plex-bootstraptest.py \
-          --destination plex \
-          --advertise-ip 127.0.0.1 \
-          --bootstrap-timeout 540 \
-          --docker-tag ${{ env.PLEX_CONTAINER_TAG }} \
-          --${{ matrix.plex }}
+      uses: nick-fields/retry@v2.8.3
+      with:
+        max_attempts: 3
+        timeout_minutes: 2
+        command: |
+          . venv/bin/activate
+          python \
+            -u tools/plex-bootstraptest.py \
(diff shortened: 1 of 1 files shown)


ReenigneArcher:
I use this action in my backport fork. Have not had a failure on the bootstrap step test since I implemented it, even with a matrix of 12 simultaneous runs.

https://github.com/LizardByte/python-plexapi-backport/blob/e792110f4ffe6dc866ae18d892b8c4c6919ec324/.github/workflows/ci.yml#L368-L388

Hope it helps!

JonnyWong16:
Any way to check/retry this?

The server bootstraps fine and libraries are set up, but it seems tests can't find the items. This seems to occur semi-regularly.

https://github.com/pkkid/python-plexapi/actions/runs/6175927368/job/16764577457?pr=1247#step:12:800

ReenigneArcher:
@JonnyWong16 yes. I have a retry step on the pytest step as well for random failures like this.

```yml
        uses: nick-fields/retry@v2.8.3
        with:
          max_attempts: 2
          timeout_minutes: 4
          command: |
            python -m pytest \
              -rxXs \
              --cache-clear \
              --ignore=tests/test_sync.py \
              --tb=native \
              --verbose \
              --cov=plexapi \
              tests
```

I don't think any tear down is required (although I might be mistaken). Just added the ``--cache-clear` arg for pytest.

I haven't tested extensively though and still get some failures that need to be manually re-run. In this case 3 failures on attempt 1 and 9 on attempt 2. So the bootstrap server may be contaminated at that point? https://github.com/LizardByte/python-plexapi-backport/actions/runs/6133253622/job/16764967441?pr=15

JonnyWong16:
Wouldn't that just keep retrying when test legitimately fail? And not just some transient failure.

ReenigneArcher:
Yes... that's why I only set it to 2 attempts. It'd probably be better to try to find the reason for the failure and try to account for it in the code. If the test is failing randomly, it's probably likely that it also fails randomly in a real environment?

I haven't looked into it too much, but maybe the item is just not in the library yet? Is it already checking if the library is refreshing before checking for an item?

JonnyWong16:
Probably just leave it alone then. I can just manually retry it.

PR #1243, opened 2023-09-11
Feature: Expose total_size in fetchItems()

glensc:
## Description

Add a way to get total for the search:

```py
maxresults = section.searchEpisodes(maxresults=0)
total_size = maxresults.total_size
```
- https://github.com/Taxel/PlexTraktSync/pull/1578

## Type of change

## Checklist:

### DIFF
--- a/plexapi/base.py
+++ b/plexapi/base.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 import re
 import weakref
+from collections import UserList
 from functools import cached_property
 from urllib.parse import urlencode
 from xml.etree import ElementTree
@@ -244,7 +245,7 @@ class PlexObject:
         if maxresults is not None:
             container_size = min(container_size, maxresults)
 
-        results = []
+        results = UserList()
         headers = {}
 
         while True:
--- a/plexapi/base.py
+++ b/plexapi/base.py
@@ -245,7 +245,6 @@ class PlexObject:
             container_size = min(container_size, maxresults)
 
         results = []
-        subresults = []
         headers = {}
 
         while True:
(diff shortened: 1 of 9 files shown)


glensc:
Maybe PlexAPI way of doing things is to add `returnTotalSize=True` kwarg instead?

glensc:
@JonnyWong16 tests pass now. but I think this change is too invasive. maybe even has small performance impact.

but any other recommendation on how to get container size? with optimal way? basically need to send query with limit=0, so it 
 (pms) would only report totalSize.

what you think of refactoring those parts of using yield rather array merging? as that's what I'm trying to solve initially:
- https://github.com/Taxel/PlexTraktSync/pull/1578

JonnyWong16:
Honestly, the public method `LibrarySection.totalViewSize(libtype="episode")` is preferred.

If you really need this then best to keep it simple and just add something like a `self._fetchTotalSize` attribute to the base object as a cache. Then you need to be very careful when you access the attribute because it will update every time `fetchItems` is called.

glensc:
I was confused by your statement:
> Honestly, the public method LibrarySection.totalViewSize(libtype="episode") is preferred.

seems `totalViewSize` method already exists:
- https://github.com/pkkid/python-plexapi/blob/7da49eb711663f8b4e2cebcac41298c6324c1427/plexapi/library.py#L475

glensc:
existing method works:
- [Taxel/PlexTraktSync@`ef93d07` (#1578)](https://github.com/Taxel/PlexTraktSync/pull/1578/commits/ef93d07206400f7b1b2b60e0e42d6ea6fd10af8a)

this mr is not needed.

PR #1244, opened 2023-09-12
Fix invalid escape sequence DeprecationWarning

cdce8p:
## Description

```
  /.../plexapi/base.py:157: DeprecationWarning: invalid escape sequence '\/'
    """ Load the specified key to find and build all items with the specified tag
```

### DIFF
--- a/plexapi/base.py
+++ b/plexapi/base.py
@@ -227,7 +227,7 @@ class PlexObject:
 
                     fetchItem(ekey, viewCount__gte=0)
                     fetchItem(ekey, Media__container__in=["mp4", "mkv"])
-                    fetchItem(ekey, guid__iregex=r"(imdb:\/\/|themoviedb:\/\/)")
+                    fetchItem(ekey, guid__iregex=r"(imdb:\\/\\/|themoviedb:\\/\\/)")
                     fetchItem(ekey, Media__Part__file__startswith="D:\\Movies")
 
         """
--- a/plexapi/base.py
+++ b/plexapi/base.py
@@ -227,7 +227,7 @@ class PlexObject:
 
                     fetchItem(ekey, viewCount__gte=0)
                     fetchItem(ekey, Media__container__in=["mp4", "mkv"])
-                    fetchItem(ekey, guid__iregex=r"(imdb:\\/\\/|themoviedb:\\/\\/)")
+                    fetchItem(ekey, guid__iregex=r"(imdb://|themoviedb://)")
                     fetchItem(ekey, Media__Part__file__startswith="D:\\Movies")
 
         """


JonnyWong16:
I'm not sure where you are seeing the `DeprecationWarning`, but that string is part of an example in the doc string and `r"(imdb:\/\/|themoviedb:\/\/)"` is valid. Technically, it doesn't even need to be escaped for Python `r"(imdb://|themoviedb://)"`.

```python
>>> import re
>>> guid = "com.plexapp.imdb://tt12345678"
>>> pattern1 = r"(imdb:\/\/|themoviedb:\/\/)"
>>> pattern2 = r"(imdb:\\/\\/|themoviedb:\\/\\/)"
>>> pattern3 = r"(imdb://|themoviedb://)"
>>> re.search(pattern1, guid)
<re.Match object; span=(12, 19), match='imdb://'>
>>> re.search(pattern2, guid)
>>> re.search(pattern3, guid)
<re.Match object; span=(12, 19), match='imdb://'>
```

Your escape pattern doesn't work.

cdce8p:
> I'm not sure where you are seeing the `DeprecationWarning`,

I got it during a pytest run with `--collection-only` surprisingly enough.

> [...] that string is part of an example in the doc string and `r"(imdb:\/\/|themoviedb:\/\/)"` is valid.

I know. When I first looked at it, I was convinced that it was a CPython bug. I even went as far as writing a bug report but stopping short of submitting it.

As I found out later, the `r` **inside** the docstring doesn't matter. It's how the complete docstring is evaluated that's important. So the escape is necessary for the docstring. Inside the documentation, it would still show it without the escape and as rawstring. (Build the docs just to make sure.)

You can try this short example if you like
```py
# test.py
def func(regex):
    """Some long comment with a rst code block

        .. code-block:: python
            func(regex=r"some\.name")
    """
```
```bash
python -X dev test.py
```

plexapi/base.py
```diff
@@ -227,7 +227,7 @@ def fetchItems(self, ekey, cls=None, container_start=None, container_size=None,
 
                     fetchItem(ekey, viewCount__gte=0)
                     fetchItem(ekey, Media__container__in=["mp4", "mkv"])
-                    fetchItem(ekey, guid__iregex=r"(imdb:\/\/|themoviedb:\/\/)")
+                    fetchItem(ekey, guid__iregex=r"(imdb:\\/\\/|themoviedb:\\/\\/)")
```
glensc, line 230:
```suggestion
                    fetchItem(ekey, guid__iregex=r"(?:imdb|themoviedb)://")
```

imho you don't need to escape `/` at all. this also optimizes regex making it more readable and skips group capture.

however, I have not tested does this actually work.

JonnyWong16: I agree. Just remove the escape characters.

PR #1251, opened 2023-09-15
Change Episode parent* attributes to properties

JonnyWong16:
## Description

Change Episode parent* attributes to properties. This is to avoid the problems with auto-reloading when Episode objects are built with seasons hidden and only reloaded when accessed.

## Type of change

## Checklist:

### DIFF
--- a/plexapi/video.py
+++ b/plexapi/video.py
@@ -886,7 +886,6 @@ class Episode(
         """ Load attribute values from Plex XML response. """
         Video._loadData(self, data)
         Playable._loadData(self, data)
-        self._seasonNumber = None  # cached season number
         self.audienceRating = utils.cast(float, data.attrib.get('audienceRating'))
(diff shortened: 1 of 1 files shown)


plexapi/video.py
```diff
@@ -925,15 +922,38 @@ def _loadData(self, data):
...
+    @cached_property
+    def parentThumb(self):
+        """ Returns the parentThumb. Refer to the Episode attributes. """
+        return self._parentThumb or self._season.thumb
+
+    @cached_property
+    def _season(self):
+        """ Returns the :class:`~plexapi.video.Season` object by querying for the show's children. """
+        return self.fetchItem(
+            f'{self.grandparentKey}/children?excludeAllLeaves=1&index={self.parentIndex}'
```
twolaw, line 955: It seems this request returns _500 internal error_ when `index=None`
See https://github.com/Taxel/PlexTraktSync/issues/1631

PR #1203, opened 2023-08-02
Improve parsing to datetime and add error handling

JonnyWong16:
## Description

* Improve parsing `utils.toDatetime()` and add error handling.
* Fix `utils.millisecondToHumanstr()` for values greater than days and negative values.

## Type of change

Please delete options that are not relevant.

- [x] New feature (non-breaking change which adds functionality)

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have added or updated the docstring for new or existing methods
- [x] I have added tests when applicable

PR #1262, opened 2023-10-02
Fix datetime returning UTC

cdce8p:
## Description
`datetime.utcfromtimestamp` will emit a DeprecationWarning in Python 3.12.

The "official" replacement would be to use `fromtimestamp(tz=timezone.utc).replace(tzinfo=None)`. However in this particular case `datetime(1970, 1, 1)` seems to be the easier solution.

https://docs.python.org/3.12/library/datetime.html#datetime.datetime.utcfromtimestamp

### DIFF
--- a/plexapi/utils.py
+++ b/plexapi/utils.py
@@ -323,7 +323,7 @@ def toDatetime(value, format=None):
                 return None
         else:
             try:
-                return datetime.utcfromtimestamp(0) + timedelta(seconds=int(value))
+                return datetime(1970, 1, 1) + timedelta(seconds=int(value))
             except ValueError:
                 log.info('Failed to parse "%s" to datetime as timestamp, defaulting to None', value)
                 return None
--- a/plexapi/utils.py
+++ b/plexapi/utils.py
@@ -323,13 +323,18 @@ def toDatetime(value, format=None):
                 return None
         else:
             try:
-                return datetime.fromtimestamp(0) + timedelta(seconds=int(value))
+                value = int(value)
             except ValueError:
                 log.info('Failed to parse "%s" to datetime as timestamp, defaulting to None', value)
                 return None
-            except OverflowError:
-                log.info('Failed to parse "%s" to datetime as timestamp (out-of-bounds), defaulting to None', value)
-                return None
+            try:
+                return datetime.fromtimestamp(value)
+            except (OSError, OverflowError):
+                try:
+                    return datetime.fromtimestamp(0) + timedelta(seconds=value)
+                except OverflowError:
+                    log.info('Failed to parse "%s" to datetime as timestamp (out-of-bounds), defaulting to None', value)
+                    return None
     return value
 
 
--- a/plexapi/utils.py
+++ b/plexapi/utils.py
@@ -323,7 +323,7 @@ def toDatetime(value, format=None):
                 return None
         else:
             try:
-                return datetime(1970, 1, 1) + timedelta(seconds=int(value))
+                return datetime.fromtimestamp(0) + timedelta(seconds=int(value))
             except ValueError:
                 log.info('Failed to parse "%s" to datetime as timestamp, defaulting to None', value)
                 return None


plexapi/utils.py
```diff
@@ -323,7 +323,7 @@ def toDatetime(value, format=None):
                 return None
         else:
             try:
-                return datetime.utcfromtimestamp(0) + timedelta(seconds=int(value))
+                return datetime(1970, 1, 1) + timedelta(seconds=int(value))
```
JonnyWong16, line 326:
```suggestion
                return datetime.fromtimestamp(0) + timedelta(seconds=int(value))
```

See https://github.com/pkkid/python-plexapi/pull/1203#issuecomment-1745062063

JonnyWong16 (changes requested):

twolaw:
Doesn't seem to be the same either.

```python
>>> datetime.datetime.utcfromtimestamp(0) + datetime.timedelta(seconds=1593721093)
datetime.datetime(2020, 7, 2, 20, 18, 13)
```

```python
>>> datetime.datetime.fromtimestamp(1593721093)
datetime.datetime(2020, 7, 2, 22, 18, 13)
```

```python
>>> datetime.datetime.fromtimestamp(0) + datetime.timedelta(seconds=1593721093)
datetime.datetime(2020, 7, 2, 21, 18, 13)
```

plexapi/utils.py
```diff
@@ -323,7 +323,7 @@ def toDatetime(value, format=None):
                 return None
         else:
             try:
-                return datetime.utcfromtimestamp(0) + timedelta(seconds=int(value))
+                return datetime.fromtimestamp(0) + timedelta(seconds=int(value))
```
twolaw, line 326: Here you calculate timezone offset from a winter date (1970-01-01) and add timedelta.
Therefore this will return a wrong time for a summer `value`.

JonnyWong16: The only thing I can think of is to use `datetime.fromtimestamp(value)` and catch `OSError` or `OverflowError` to fallback to `datetime.fromtimestamp(0) + timedelta(seconds=int(value)` if the `value` is negative or very large.

twolaw: I don't what was wrong with the previous code :
```python
value = min(max(int(value), 86400), 2**31 - 1)
value = datetime.fromtimestamp(int(value))
```
We seem to agree with the `datetime.fromtimestamp(value)` part.
Remains the handling of offlimit values.

JonnyWong16: Right, the same as previous code, but instead of the hacky max/min, just catch the exception and use the `timedelta` from 0.

For normal values it would take into account the timezone, but out-of-bounds values won't use the correct timezone. If the values are out-of-bounds then something else has probably gone wrong and a small timezone error is probably the least of your worries.

cdce8p: > Right, the same as previous code, but instead of the hacky max/min, just catch the exception and use the `timedelta` from 0.

Should I change anything else? AFAICT this is what's implemented atm with `except OverflowError`.

JonnyWong16:
I think that should work now.

JonnyWong16 (approved):

meisnate12:
I've got a user getting an issue that i think may be related to this PR. The plex instance seems to have some weird addedAt dates and causing `ValueError: year 30828 is out of range` to be thrown. I'm trying to get the user to fix the metadata in their DB, but they mentioned that it was working fine on the latest PMM which uses PlexAPI v4.15.2 but when switched to develop using v4.15.6 he would see this issue. I can open a new issue if needed but thought id post on this PR first.
```
LibrarySection.findItems(data, initpath=key)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/usr/local/lib/python3.11/site-packages/plexapi/base.py", line 329, in findItems
  item = self._buildItemOrNone(elem, cls, initpath)
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/usr/local/lib/python3.11/site-packages/plexapi/base.py", line 109, in _buildItemOrNone
  return self._buildItem(elem, cls, initpath)
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/usr/local/lib/python3.11/site-packages/plexapi/base.py", line 101, in _buildItem
  return ecls(self._server, elem, initpath)
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/usr/local/lib/python3.11/site-packages/plexapi/base.py", line 62, in __init__
  self._loadData(data)
File "/usr/local/lib/python3.11/site-packages/plexapi/video.py", line 363, in _loadData
  Video._loadData(self, data)
File "/usr/local/lib/python3.11/site-packages/plexapi/video.py", line 51, in _loadData
  self.addedAt = utils.toDatetime(data.attrib.get('addedAt'))
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/usr/local/lib/python3.11/site-packages/plexapi/utils.py", line 331, in toDatetime
  return datetime.fromtimestamp(value)
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ValueError: year 30828 is out of range
```

PR #1288, opened 2023-11-06
Add sonicallySimilar method to Audio class

### LINKED ISSUE #1183: Sonic Analysis API Access

I am unsure if Plex exposes the Sonic Analysis feature of Plex to their API. Request an investigation if it does, and if so, that the python package is update to wrap these commands. Thanks!

Unknown, perhaps with XML API access.

### Code Snippets

N/A

### Additional Context

N/A

Dr-Blank:
closes #1183

## Description

This pull request adds a new method, `sonicallySimilar`, to the `Audio` class. This method allows users to find sonically similar audio items based on a maximum distance between tracks. The method takes in optional parameters for `limit` and `maxDistance` and returns a list of sonically similar audio items.

`Audio` class now contains a new attribute `distance` which is the sonic distance from the seed item.

## Type of change

## Checklist:

### DIFF
--- a/plexapi/audio.py
+++ b/plexapi/audio.py
@@ -3,7 +3,7 @@ import os
 from pathlib import Path
 from urllib.parse import quote_plus
 
-from typing import TypeVar
+from typing import List, Optional, TypeVar
 
 from plexapi import media, utils
 from plexapi.base import Playable, PlexPartialObject, PlexHistory, PlexSession
@@ -134,22 +134,27 @@ class Audio(PlexPartialObject, PlayedUnplayedMixin):
 
     def sonicallySimilar(
         self: TAudio,
-        limit: int = 30,
-        maxDistance: float = 0.25,
+        limit: Optional[int] = None,
+        maxDistance: Optional[float] = None,
         **kwargs,
     ) -> List[TAudio]:
         """Returns a list of sonically similar audio items.
 
         Parameters:
-            limit (int): maximum count of items to return, unlimited if `None`.
-            maxDistance (float): maximum distance between tracks, 0.0 - 1.0.
+            limit (int): Maximum count of items to return. Default 50 (server default)
+            maxDistance (float): Maximum distance between tracks, 0.0 - 1.0. Default 0.25 (server default).
             **kwargs: Additional options passed into :func:`~plexapi.base.PlexObject.fetchItems`.
 
         Returns:
             List[:class:`~plexapi.audio.Audio`]: list of sonically similar audio items.
         """
 
-        key = f"/library/metadata/{self.ratingKey}/nearest?limit={limit}&maxDistance={maxDistance}"
+        key = f"{self.key}/nearest"
+        params = {"maxDistance": maxDistance, "limit": limit}
+        key += utils.joinArgs(
+            {k: v for k, v in params.items() if v is not None}
+        )
+
         return self.fetchItems(
             key,
             cls=self.__class__,
--- a/plexapi/audio.py
+++ b/plexapi/audio.py
@@ -3,7 +3,7 @@ import os
 from pathlib import Path
 from urllib.parse import quote_plus
 
-from typing import List, Optional, TypeVar
+from typing import Any, Dict, List, Optional, TypeVar
 
 from plexapi import media, utils
 from plexapi.base import Playable, PlexPartialObject, PlexHistory, PlexSession
@@ -59,6 +59,7 @@ class Audio(PlexPartialObject, PlayedUnplayedMixin):
         self.addedAt = utils.toDatetime(data.attrib.get('addedAt'))
         self.art = data.attrib.get('art')
         self.artBlurHash = data.attrib.get('artBlurHash')
+        self.distance = utils.cast(float, data.attrib.get('distance'))
         self.fields = self.findItems(data, media.Field)
         self.guid = data.attrib.get('guid')
         self.index = utils.cast(int, data.attrib.get('index'))
@@ -71,7 +72,6 @@ class Audio(PlexPartialObject, PlayedUnplayedMixin):
         self.listType = 'audio'
         self.moods = self.findItems(data, media.Mood)
         self.musicAnalysisVersion = utils.cast(int, data.attrib.get('musicAnalysisVersion'))
-        self.distance = utils.cast(float, data.attrib.get('distance'))
         self.ratingKey = utils.cast(int, data.attrib.get('ratingKey'))
         self.summary = data.attrib.get('summary')
         self.thumb = data.attrib.get('thumb')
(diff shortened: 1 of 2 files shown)


Dr-Blank (commented):
@JonnyWong16 What is recommendation here, give up on typehinting by removing the import or add a dependency? or is there another way of hinting self in python 3.8 that I am missing?

`-> "list[Audio]"` does not serve the same purpose as `-> "list[Self]"` in this case

JonnyWong16:
You can use a `TypeVar`.

https://realpython.com/python-type-self/#annotating-with-typevar

`from __future__ import annotations` won't work well here because of the subclassing of `Audio`.

plexapi/audio.py
```diff
@@ -14,6 +16,9 @@
 from plexapi.playlist import Playlist
 
 
+Self = TypeVar("Self", bound="Audio")
```
JonnyWong16, line 19:
```suggestion
TAudio = TypeVar("TAudio", bound="Audio")
```

plexapi/audio.py
```diff
@@ -3,6 +3,8 @@
 from pathlib import Path
 from urllib.parse import quote_plus
 
+from typing import TypeVar
```
JonnyWong16, line 6:
```suggestion
from typing import List, TypeVar
```

plexapi/audio.py
```diff
@@ -125,6 +132,30 @@ def sync(self, bitrate, client=None, clientId=None, limit=None, title=None):
 
         return myplex.sync(sync_item, client=client, clientId=clientId)
 
+    def sonicallySimilar(
+        self: Self,
```
JonnyWong16, line 136:
```suggestion
        self: TAudio,
```

plexapi/audio.py
```diff
@@ -125,6 +132,30 @@ def sync(self, bitrate, client=None, clientId=None, limit=None, title=None):
...
+        limit: int = 30,
+        maxDistance: float = 0.25,
+        **kwargs,
+    ) -> "list[Self]":
```
JonnyWong16, line 140:
```suggestion
    ) -> List[TAudio]:
```

plexapi/audio.py
```diff
@@ -125,6 +132,30 @@ def sync(self, bitrate, client=None, clientId=None, limit=None, title=None):
...
+        """ Find sonically similar audio items.
```
JonnyWong16, line 141:
```suggestion
        """ Returns a list of sonically similar audio items.
```

plexapi/audio.py
```diff
@@ -34,6 +39,7 @@ class Audio(PlexPartialObject, PlayedUnplayedMixin):
             listType (str): Hardcoded as 'audio' (useful for search filters).
             moods (List<:class:`~plexapi.media.Mood`>): List of mood objects.
             musicAnalysisVersion (int): The Plex music analysis version for the item.
+            distance (float): Sonic Distance of the item from the seed item.
```
JonnyWong16, line 42: Attributes are listed in alphabetical order.

plexapi/audio.py
```diff
@@ -125,6 +132,30 @@ def sync(self, bitrate, client=None, clientId=None, limit=None, title=None):
...
+        Parameters:
+            limit (int): maximum count of items to return, unlimited if `None`.
+            maxDistance (float): maximum distance between tracks, 0.0 - 1.0.
+            **kwargs: Additional options passed into :func:`~plexapi.base.PlexObject.fetchItems`.
+
+        Returns:
+            List[:class:`~plexapi.audio.Audio`]: list of sonically similar audio items.
+        """
+
+        key = f"/library/metadata/{self.ratingKey}/nearest?limit={limit}&maxDistance={maxDistance}"
```
JonnyWong16, line 152:
```suggestion
        key = f"{self.key}/nearest?limit={limit}&maxDistance={maxDistance}"
```

Does `limit=None` need to be excluded from the query string?

plexapi/audio.py
```diff
@@ -125,6 +132,30 @@ def sync(self, bitrate, client=None, clientId=None, limit=None, title=None):
...
+        self: Self,
+        limit: int = 30,
+        maxDistance: float = 0.25,
+        **kwargs,
+    ) -> "list[Self]":
+        """ Find sonically similar audio items.
+
+        Parameters:
+            limit (int): maximum count of items to return, unlimited if `None`.
+            maxDistance (float): maximum distance between tracks, 0.0 - 1.0.
```
JonnyWong16, line 145:
```suggestion
            limit (int): Maximum count of items to return, unlimited if `None`. Default 30.
            maxDistance (float): Maximum distance between tracks, 0.0 - 1.0. Default 0.25.
```

JonnyWong16 (changes requested):

plexapi/audio.py
Dr-Blank: Value 'list' is unsubscriptable for py3.8, using List from typing instead

plexapi/audio.py
Dr-Blank: fixed in the latest commit

- None would result in it using the server defaults which is added in the docstring

tests/test_audio.py
```diff
@@ -439,6 +439,12 @@ def test_audio_Audio_section(artist, album, track):
     assert track.section().key == album.section().key == artist.section().key
 
 
+def test_audio_Audio_sonicallySimilar(artist):
```
Dr-Blank, line 442:
```python
    similar_audio = artist.sonicallySimilar(limit=1)
    assert len(similar_audio) < 2

    similar_audio = artist.sonicallySimilar(maxDistance=0.1)
    assert all(i.distance <= 0.1 for i in similar_audio)
```
I have purposefully not added these tests as I think they are not in the scope of this api

plexapi/audio.py
```diff
@@ -65,6 +71,7 @@ def _loadData(self, data):
         self.listType = 'audio'
         self.moods = self.findItems(data, media.Mood)
         self.musicAnalysisVersion = utils.cast(int, data.attrib.get('musicAnalysisVersion'))
+        self.distance = utils.cast(float, data.attrib.get('distance'))
```
JonnyWong16, line 74: Attributes in alphabetical order.

plexapi/audio.py
```diff
@@ -125,6 +132,35 @@ def sync(self, bitrate, client=None, clientId=None, limit=None, title=None):
...
+
+        Returns:
+            List[:class:`~plexapi.audio.Audio`]: list of sonically similar audio items.
+        """
+
+        key = f"{self.key}/nearest"
+        params = {"maxDistance": maxDistance, "limit": limit}
+        key += utils.joinArgs(
+            {k: v for k, v in params.items() if v is not None}
+        )
```
JonnyWong16, line 156: It is more clear to add params instead of removing params.
```suggestion
        params = {}
        if limit is not None:
            params['limit'] = limit
        if maxDistance is not None:
            params['maxDistance'] = maxDistance
        key += utils.joinArgs(params)
```

tests/test_audio.py
JonnyWong16: Add them so that we at least have coverage for the optional parameters. The results of the test will be meaningless.

JonnyWong16 (changes requested):

plexapi/audio.py
Dr-Blank: I have a way to maintain the adding of param as well as make it add programmatically so that it is scalable and avoid code repetition of `if` statements for every key

```py
        all_args = {"limit": limit, "maxDistance": maxDistance}
        params = {k: v for k, v in all_args.items() if v is not None}  # add if not None


```

JonnyWong16 (approved):

Issue #1296, opened 2023-11-11
Add Type Hints to Improve Codebase Readability and Maintainability


Dr-Blank:
### What is your feature request?

I suggest adding type hints throughout the codebase to enhance its clarity and maintainability. Type hints provide several benefits:

1. **Readability:** Type hints serve as documentation, making it easier for developers to understand the code and its intended usage.

2. **Early Detection of Errors:** Type hints enable static type checking tools to catch potential issues at compile-time, reducing the likelihood of runtime errors.

3. **Tooling Support:** Type hints improve IDE support, enabling better autocompletion, refactoring, and overall developer experience.

4. **Improved Collaboration:** With type hints, developers can more easily collaborate on the project, as the code becomes more self-explanatory.

5. **Reduced Debugging Time:** Type hints can help identify and resolve issues more quickly during the development process.

6. **Enhanced Code Quality:** The addition of type hints promotes better coding practices and can lead to a more robust and maintainable codebase.

### Additional Context

If this proposal resonates with the community, I am willing to contribute to the implementation of type hints. I believe this enhancement will significantly benefit both current and future contributors.

Thank you for considering this improvement.

Dr-Blank:
@JonnyWong16 

can attr `key` be None in any case? if so, for which cases and is `key=None` useful? as I believe all media items must have a key, hence key can not be None for all of those cases. correct me if I am wrong, but if that is the case, a base class for those items should make it `key` compulsory and for objects that have `key` as None should not inherit from this class with attr `key` in it.

I am saying this as it will save a ton of `if self.key is None` checks in methods which actually use `key`

PR #1302, opened 2023-11-16
add type hints to `base.py`

Dr-Blank:
## Description

Support mypy and/or pyright for the `base.py`

related #1296 

## Type of change

## Checklist:

### DIFF
--- a/plexapi/audio.py
+++ b/plexapi/audio.py
@@ -1,11 +1,13 @@
 # -*- coding: utf-8 -*-
+from __future__ import annotations
+
 import os
 from pathlib import Path
+from typing import TYPE_CHECKING
 from urllib.parse import quote_plus
-from xml.etree.ElementTree import Element
 
 from plexapi import media, utils
-from plexapi.base import Playable, PlexPartialObject, PlexHistory, PlexSession
+from plexapi.base import Playable, PlexHistory, PlexPartialObject, PlexSession
 from plexapi.exceptions import BadRequest
 from plexapi.mixins import (
     AdvancedSettingsMixin, SplitMergeMixin, UnmatchMatchMixin, ExtrasMixin, HubsMixin, PlayedUnplayedMixin, RatingMixin,
@@ -13,6 +15,9 @@ from plexapi.mixins import (
     ArtistEditMixins, AlbumEditMixins, TrackEditMixins
 )
 from plexapi.playlist import Playlist
+
+if TYPE_CHECKING:
+    from xml.etree.ElementTree import Element
 
 
 class Audio(PlexPartialObject, PlayedUnplayedMixin):
--- a/plexapi/audio.py
+++ b/plexapi/audio.py
@@ -2,6 +2,7 @@
 import os
 from pathlib import Path
 from urllib.parse import quote_plus
+from xml.etree.ElementTree import Element
 
 from plexapi import media, utils
 from plexapi.base import Playable, PlexPartialObject, PlexHistory, PlexSession
@@ -47,7 +48,7 @@ class Audio(PlexPartialObject, PlayedUnplayedMixin):
     """
     METADATA_TYPE = 'track'
 
-    def _loadData(self, data):
+    def _loadData(self, data: Element):
         """ Load attribute values from Plex XML response. """
         self._data = data
         self.addedAt = utils.toDatetime(data.attrib.get('addedAt'))
@@ -154,7 +155,7 @@ class Artist(
     TAG = 'Directory'
     TYPE = 'artist'
 
-    def _loadData(self, data):
+    def _loadData(self, data: Element):
         """ Load attribute values from Plex XML response. """
         Audio._loadData(self, data)
         self.albumSort = utils.cast(int, data.attrib.get('albumSort', '-1'))
@@ -285,7 +286,7 @@ class Album(
     TAG = 'Directory'
     TYPE = 'album'
 
-    def _loadData(self, data):
+    def _loadData(self, data: Element):
         """ Load attribute values from Plex XML response. """
         Audio._loadData(self, data)
         self.collections = self.findItems(data, media.Collection)
@@ -417,7 +418,7 @@ class Track(
     TAG = 'Track'
     TYPE = 'track'
 
-    def _loadData(self, data):
+    def _loadData(self, data: Element):
         """ Load attribute values from Plex XML response. """
         Audio._loadData(self, data)
         Playable._loadData(self, data)
@@ -497,7 +498,7 @@ class TrackSession(PlexSession, Track):
     """
     _SESSIONTYPE = True
 
-    def _loadData(self, data):
+    def _loadData(self, data: Element):
         """ Load attribute values from Plex XML response. """
         Track._loadData(self, data)
         PlexSession._loadData(self, data)
@@ -510,7 +511,7 @@ class TrackHistory(PlexHistory, Track):
     """
     _HISTORYTYPE = True
 
-    def _loadData(self, data):
+    def _loadData(self, data: Element):
         """ Load attribute values from Plex XML response. """
         Track._loadData(self, data)
         PlexHistory._loadData(self, data)
--- a/plexapi/base.py
+++ b/plexapi/base.py
@@ -17,7 +17,6 @@ if typing.TYPE_CHECKING:
     from plexapi.media import Session
     from plexapi.server import PlexServer
 
-# FunctionCheck = Callable[..., bool]
 FunctionCheck = Callable[[Any, Any], bool]
 PlexObjectT = TypeVar('PlexObjectT', bound='PlexObject')
 
@@ -184,7 +183,12 @@ class PlexObject:
     ) -> Optional[PlexObject]:
         ...
 
-    def _buildItemOrNone(self, elem: Element, cls: Optional[Type[PlexObjectT]] = None, initpath: Optional[str] = None):
+    def _buildItemOrNone(
+        self,
+        elem: Element,
+        cls: Optional[Type[PlexObjectT]] = None,
+        initpath: Optional[str] = None,
+    ):
         """ Calls :func:`~plexapi.base.PlexObject._buildItem` but returns
             None if elem is an unknown type.
         """
@@ -539,7 +543,7 @@ class PlexObject:
         self._initpath = key
         data = self._server.query(key)
         if not data:
-            raise NotFound(f'Unable to find elem: key={key}')
+            raise NotFound(f'Unable to find elem: {key=}')
         self._overwriteNone = _overwriteNone
         self._loadData(data[0])
         self._overwriteNone = True
--- a/plexapi/base.py
+++ b/plexapi/base.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 from __future__ import annotations
+from datetime import datetime
 
 import re
 from typing import Any, Callable, Dict, List, Optional, Set, Type, TypeVar, Union, cast, overload
@@ -16,6 +17,7 @@ from plexapi.exceptions import BadRequest, NotFound, UnknownType, Unsupported
 if typing.TYPE_CHECKING:
     from plexapi.media import Session
     from plexapi.server import PlexServer
+    from plexapi.library import LibrarySection
 
 FunctionCheck = Callable[[Any, Any], bool]
 PlexObjectT = TypeVar('PlexObjectT', bound='PlexObject')
@@ -95,7 +97,7 @@ class PlexObject:
         # Automatically reload the object when accessing a missing attribute
         self._autoReload = CONFIG.get('plexapi.autoreload', True, bool)
         # Attribute to save batch edits for a single API call
-        self._edits = None
+        self._edits: Optional[Dict[str, Any]] = None
 
         if data is not None:
             self._loadData(data)
@@ -709,7 +711,7 @@ class PlexPartialObject(PlexObject):
         self.section()._edit(items=self, **kwargs)
         return self
 
-    def edit(self, **kwargs):
+    def edit(self, **kwargs: Any):
         """ Edit an object.
             Note: This is a low level method and you need to know all the field/tag keys.
             See :class:`~plexapi.mixins.EditFieldMixin` and :class:`~plexapi.mixins.EditTagsMixin`
@@ -737,7 +739,7 @@ class PlexPartialObject(PlexObject):
         """
         return self._edit(**kwargs)
 
-    def batchEdits(self):
+    def batchEdits(self) -> PlexPartialObject:
         """ Enable batch editing mode to save API calls.
             Must call :func:`~plexapi.base.PlexPartialObject.saveEdits` at the end to save all the edits.
             See :class:`~plexapi.mixins.EditFieldMixin` and :class:`~plexapi.mixins.EditTagsMixin`
@@ -788,8 +790,10 @@ class PlexPartialObject(PlexObject):
         key = f'{self.key}/refresh'
         self._server.query(key, method=self._server._session.put)
 
-    def section(self):
+    def section(self) -> LibrarySection:
         """ Returns the :class:`~plexapi.library.LibrarySection` this item belongs to. """
+        # TODO: consider moving this to MediaContainer as `librarySectionByID` is not available
+        #       or bring in the attr to this class
         return self._server.library.sectionByID(self.librarySectionID)
 
     def delete(self):
@@ -801,7 +805,7 @@ class PlexPartialObject(PlexObject):
                 'have not allowed items to be deleted', self.key)
             raise
 
-    def history(self, maxresults=None, mindate=None):
+    def history(self, maxresults: Optional[int] = None, mindate: Optional[datetime] = None):
         """ Get Play History for a media item.
 
             Parameters:
--- a/plexapi/base.py
+++ b/plexapi/base.py
@@ -1,11 +1,11 @@
 # -*- coding: utf-8 -*-
 from __future__ import annotations
-from datetime import datetime
 
 import re
-from typing import Any, Callable, Dict, List, Optional, Set, Type, TypeVar, Union, cast, overload, TYPE_CHECKING
 import weakref
 from functools import cached_property
+from typing import (TYPE_CHECKING, Any, Callable, Dict, List, Optional, Set,
+                    Type, TypeVar, Union, cast, overload)
 from urllib.parse import urlencode
 from xml.etree import ElementTree
 from xml.etree.ElementTree import Element
@@ -14,9 +14,11 @@ from plexapi import CONFIG, X_PLEX_CONTAINER_SIZE, log, utils
 from plexapi.exceptions import BadRequest, NotFound, UnknownType, Unsupported
 
 if TYPE_CHECKING:
+    from datetime import datetime
+
+    from plexapi.library import LibrarySection
     from plexapi.media import Session
     from plexapi.server import PlexServer
-    from plexapi.library import LibrarySection
 
 FunctionCheck = Callable[[Any, Any], bool]
 PlexObjectT = TypeVar('PlexObjectT', bound='PlexObject')
--- a/plexapi/base.py
+++ b/plexapi/base.py
@@ -3,8 +3,7 @@ from __future__ import annotations
 from datetime import datetime
 
 import re
-from typing import Any, Callable, Dict, List, Optional, Set, Type, TypeVar, Union, cast, overload
-import typing
+from typing import Any, Callable, Dict, List, Optional, Set, Type, TypeVar, Union, cast, overload, TYPE_CHECKING
 import weakref
 from functools import cached_property
 from urllib.parse import urlencode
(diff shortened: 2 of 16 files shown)


plexapi/audio.py
```diff
@@ -2,6 +2,7 @@
 import os
 from pathlib import Path
 from urllib.parse import quote_plus
+from xml.etree.ElementTree import Element
```
glensc, line 5: perhaps add such imports that only used for typing under TYPE_CHECKING check?

example:
- https://github.com/Taxel/PlexTraktSync/blob/2f1235e127d69491f3d748d54eccdec9504aef62/plextraktsync/queue/Queue.py#L5-L8

this gives back small performance loss when importing more modules

plexapi/base.py
```diff
@@ -1,33 +1,48 @@
 # -*- coding: utf-8 -*-
+from __future__ import annotations
```
glensc, line 2: for consistency, imho this must be added to any file that uses typing

plexapi/base.py
```diff
@@ -1,33 +1,48 @@
...
+from datetime import datetime
+
 import re
+from typing import Any, Callable, Dict, List, Optional, Set, Type, TypeVar, Union, cast, overload
+import typing
```
glensc, line 7: double imports:
- from typing ...
- import typing

plexapi/base.py
```diff
@@ -321,17 +464,24 @@ def findItems(self, data, cls=None, initpath=None, rtag=None, **kwargs):
             kwargs['type'] = cls.TYPE
         # rtag to iter on a specific root tag using breadth-first search
         if rtag:
-            data = next(utils.iterXMLBFS(data, rtag), [])
+            data = next(utils.iterXMLBFS(data, rtag), Element('empty'))
         # loop through all data elements to find matches
-        items = []
+        items: List[Union[PlexObject, PlexObjectT]] = []
```
glensc, line 469:
```suggestion
        items: list[Union[PlexObject, PlexObjectT]] = []
```

afaik can use native list

plexapi/audio.py
JonnyWong16: In this case, this should have no impact on performance because `Element` is already imported in a different module.

plexapi/base.py
JonnyWong16: Not in Python 3.8.

plexapi/base.py
Dr-Blank: I am planning to typehint other files as well and will implement this for their respective PRs if thats fine.

Dr-Blank:
Is there any task remaining for this PR?

plexapi/base.py
```diff
@@ -110,12 +198,12 @@ def _buildItemOrNone(self, elem, cls=None, initpath=None):
         except UnknownType:
             return None
 
-    def _buildDetailsKey(self, **kwargs):
+    def _buildDetailsKey(self, **kwargs: Any):
```
ReenigneArcher, line 201:
```suggestion
    def _buildDetailsKey(self, **kwargs: Optional[Dict[str, Any]]):
```

Shouldn't kwargs be a dictionary?

I'm unsure the type of the value of the dict, so left that as `Any`

ReenigneArcher (commented):
I have some comments/suggestions.

Some of my comments apply in multiple places, but I only made the comment once. For example the kwargs type hint comment.

Note, I am not a maintainer of this project, but a contributor and user of it.

plexapi/base.py
Dr-Blank: is a shorthand, `**_` is automatically parsed as `Dict[str, Unknown]`. providing `Any` is for value as key and dict is understood

plexapi/playlist.py
```diff
@@ -2,9 +2,10 @@
 import re
 from pathlib import Path
 from urllib.parse import quote_plus, unquote
+from xml.etree.ElementTree import Element
 
 from plexapi import media, utils
-from plexapi.base import Playable, PlexPartialObject
+from plexapi.base import Playable
```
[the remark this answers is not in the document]
Dr-Blank: maybe i should change my PR title to something like support `base.py` for mypy and pyright. as this change is for that, plexobject is changed as those attributes are used in that class in its methods but are defined in its children. other option would be to move those methods to children.

plexapi/base.py
```diff
@@ -40,23 +56,49 @@ class PlexObject:
             initpath (str): Relative path requested when retrieving specified `data` (optional).
             parent (:class:`~plexapi.base.PlexObject`): The parent object that this object is built from (optional).
     """
-    TAG = None      # xml element tag
-    TYPE = None     # xml element type
-    key = None      # plex relative url
+    TAG: Optional[str] = None  # xml element tag
+    TYPE: Optional[str] = None  # xml element type
+    key: Optional[str] = None  # plex relative url
+    _INCLUDES = {
```
JonnyWong16, line 62: Moving `_INCLUDES` here causes them to be added to all `PlexObject` in `_buildDetailsKey`. The includes should only be added to the details key for `PlexPartialObjects`.

plexapi/media.py
```diff
@@ -1,13 +1,20 @@
...
 
 from plexapi import log, settings, utils
 from plexapi.base import PlexObject
 from plexapi.exceptions import BadRequest
 from plexapi.utils import deprecated
 
+if TYPE_CHECKING:
+    from xml.etree.ElementTree import Element
+
+
```
JonnyWong16, line 17: Remove extra blank line.

JonnyWong16 (changes requested):
I have only done a quick review on GitHub. I haven't actually checked out this PR in my code editor to check.

JonnyWong16:
I also created a new branch for `typehints`. I think it would be better to finish type hinting the entire library before merging into `master.

plexapi/base.py
```diff
@@ -1,33 +1,49 @@
 # -*- coding: utf-8 -*-
+from __future__ import annotations
+
 import re
 import weakref
 from functools import cached_property
+from typing import (TYPE_CHECKING, Any, Callable, Dict, List, Optional, Set,
+                    Type, TypeVar, Union, cast, overload)
```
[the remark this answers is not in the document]
Dr-Blank: default config or have something in mind for this project?

- all single quotes will be converted to double unless "--skip-string-normalization" is passed
- all docstrings whitespaces will be stripped

plexapi/base.py
```diff
@@ -82,26 +124,74 @@ def _clean(self, value):
...
+        initpath: Optional[str] = None,
+    ) -> Optional[PlexObject]:
+        ...
+
+    def _buildItemOrNone(
+        self,
+        elem: Element,
+        cls: Optional[Type[PlexObjectT]] = None,
+        initpath: Optional[str] = None,
+    ):
```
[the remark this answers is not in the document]
Dr-Blank: yes, Optional[Union[PlexObjectT, PlexObject]]

plexapi/base.py
```diff
@@ -110,12 +200,12 @@ def _buildItemOrNone(self, elem, cls=None, initpath=None):
...
             return None
 
-    def _buildDetailsKey(self, **kwargs):
+    def _buildDetailsKey(self, **kwargs: Any):
         """ Builds the details key with the XML include parameters.
             All parameters are included by default with the option to override each parameter
             or disable each parameter individually by setting it to False or 0.
         """
-        details_key = self.key
+        details_key = str(self.key)
```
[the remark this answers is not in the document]
Dr-Blank: good catch, removed casting

plexapi/base.py
```diff
@@ -126,22 +216,38 @@ def _buildDetailsKey(self, **kwargs):
...
+    def _isChildOf(self, **kwargs: Any):
         """ Returns True if this object is a child of the given attributes.
             This will search the parent objects all the way to the top.
 
             Parameters:
                 **kwargs (dict): The attributes and values to search for in the parent objects.
                     See all possible `**kwargs*` in :func:`~plexapi.base.PlexObject.fetchItem`.
         """
-        obj = self
+        obj = self  # ? Why is this needed?
```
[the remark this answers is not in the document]
Dr-Blank: isn't object address copied here? or am i missing something?
```python
class Car:
    def __init__(self, name: str, color: str) -> None:
        self.name = name
        self.color = color

    def _dont_mutate_me(self) -> None:
        obj = self  # only memory address of self is copied not the object
        obj.name = "foo"  # will mutate self.name
        assert id(obj) == id(self)  # True

    def __str__(self) -> str:
        return f"I am a {self.color} {self.name}."


def main() -> None:
    car = Car(name="Tesla", color="red")
    print(car)  # I am a red Tesla.
    car._dont_mutate_me()
    print(car)  # I am a red foo.


if __name__ == "__main__":
    main()
```

plexapi/base.py
```diff
@@ -40,23 +56,49 @@ class PlexObject:
...
+        'includeStations': 1
+    }
 
-    def __init__(self, server, data, initpath=None, parent=None):
+    def __init__(
+        self,
+        server: PlexServer,
+        data: Element,
+        initpath: Optional[str] = None,
+        parent: Optional[PlexObject] = None,
```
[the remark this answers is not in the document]
Dr-Blank: mypy complains
`base.py:167: error: Argument "parent" to "PlexObject" has incompatible type "PlexObject"; expected "PlexObjectT | None"  [arg-type]`

not sure why

was thinking no need for generic type as we are not using that generic type again in any other type definition for the function

plexapi/base.py
```diff
@@ -126,22 +216,38 @@ def _buildDetailsKey(self, **kwargs):
...
+
+    @overload
+    def _manuallyLoadXML(
+        self,
+        xml: str,
+        cls: None = None,
+    ) -> PlexObject:
+        ...
+
+    def _manuallyLoadXML(self, xml: str, cls: Optional[Type[PlexObjectT]] = None):
```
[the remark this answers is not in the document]
Dr-Blank: updated to `Optional[Union[PlexObjectT, PlexObject]]`

plexapi/base.py
```diff
@@ -424,22 +576,22 @@ def _checkAttrs(self, elem, **kwargs):
...
+    def _getAttrValue(self, elem: Element, attrstr: str, results: Optional[List[str]] = None) -> List[str]:
         # log.debug('Fetching %s in %s', attrstr, elem.tag)
         parts = attrstr.split('__', 1)
         attr = parts[0]
-        attrstr = parts[1] if len(parts) == 2 else None
+        attrstr = parts[1] if len(parts) == 2 else ""
         if attrstr:
             results = [] if results is None else results
-            for child in [c for c in elem if c.tag.lower() == attr.lower()]:
+            for child in filter(lambda c: c.tag.lower() == attr.lower(), elem):
```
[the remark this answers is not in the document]
Dr-Blank: not related to typehints, will remove this from the PR

performance improvment as filter returns an iterator and yields value lazily as opposed to creating the full list in memory.

plexapi/base.py
Dr-Blank: that is something i did not know. fixed by including an empty member on `PlexObject` to satisfy pyright

plexapi/base.py
```diff
@@ -1054,7 +1067,7 @@ def _reload(self, _autoReload=False, **kwargs):
 
     def source(self):
         """ Return the source media object for the session. """
-        return self.fetchItem(self._details_key)
+        return self._details_key and self.fetchItem(self._details_key)
```
Dr-Blank, line 1102: fetchItems does not accept None

Dr-Blank (commented):
marked side effect of suggestion

plexapi/base.py
```diff
@@ -231,10 +337,8 @@ def fetchItems(self, ekey, cls=None, container_start=None, container_size=None,
                     fetchItem(ekey, Media__Part__file__startswith="D:\\Movies")
 
         """
-        if ekey is None:
-            raise BadRequest('ekey was not provided')
```
JonnyWong16, line 235: Why is guard clause removed?

plexapi/base.py
```diff
@@ -253,7 +357,9 @@ def fetchItems(self, ekey, cls=None, container_start=None, container_size=None,
             headers['X-Plex-Container-Size'] = str(container_size)
 
             data = self._server.query(ekey, headers=headers)
-            subresults = self.findItems(data, cls, ekey, **kwargs)
+            if not data:
+                return []
```
JonnyWong16, line 361: Is this necessary?

plexapi/base.py
```diff
@@ -393,20 +535,22 @@ def reload(self, key=None, **kwargs):
...
+    def _reload(self, key: Optional[str] = None, _overwriteNone: bool = True, **kwargs: Any):
         """ Perform the actual reload. """
         details_key = self._buildDetailsKey(**kwargs) if kwargs else self._details_key
         key = key or details_key or self.key
         if not key:
             raise Unsupported('Cannot reload an object not built from a URL.')
         self._initpath = key
         data = self._server.query(key)
+        if data is None:
+            raise NotFound(f'Unable to find elem: {key=}')
```
JonnyWong16, line 547: Is this necessary?

plexapi/base.py
```diff
@@ -40,23 +68,29 @@ class PlexObject:
...
+        server: PlexServer,
+        data: Element,
+        initpath: Optional[str] = None,
+        parent: Optional[PlexObject] = None,
+    ):
         self._server = server
         self._data = data
         self._initpath = initpath or self.key
         self._parent = weakref.ref(parent) if parent is not None else None
-        self._details_key = None
```
JonnyWong16, line 52: Why is this removed?

plexapi/base.py
```diff
@@ -940,6 +1089,8 @@ def _reload(self, _autoReload=False, **kwargs):
 
         key = self._initpath
         data = self._server.query(key)
+        if not data:
+            raise NotFound(f'Unable to find elem: {key=}')
```
JonnyWong16, line 1093: Is this necessary?

plexapi/collection.py
```diff
@@ -1,18 +1,22 @@
...
 from plexapi.exceptions import BadRequest, NotFound, Unsupported
 from plexapi.library import LibrarySection, ManagedHub
-from plexapi.mixins import (
-    AdvancedSettingsMixin, SmartFilterMixin, HubsMixin, RatingMixin,
-    ArtMixin, PosterMixin, ThemeMixin,
-    CollectionEditMixins
-)
+from plexapi.mixins import (AdvancedSettingsMixin, ArtMixin,
+                            CollectionEditMixins, HubsMixin, PosterMixin,
+                            RatingMixin, SmartFilterMixin, ThemeMixin)
```
JonnyWong16, line 14: Don't change in this PR.

plexapi/library.py
```diff
@@ -1,19 +1,25 @@
...
 from plexapi.base import OPERATORS, PlexObject
 from plexapi.exceptions import BadRequest, NotFound
-from plexapi.mixins import (
-    MovieEditMixins, ShowEditMixins, SeasonEditMixins, EpisodeEditMixins,
-    ArtistEditMixins, AlbumEditMixins, TrackEditMixins, PhotoalbumEditMixins, PhotoEditMixins
-)
+from plexapi.mixins import (AlbumEditMixins, ArtistEditMixins,
+                            EpisodeEditMixins, MovieEditMixins,
+                            PhotoalbumEditMixins, PhotoEditMixins,
+                            SeasonEditMixins, ShowEditMixins, TrackEditMixins)
```
JonnyWong16, line 16: Don't change in this PR.

plexapi/photo.py
```diff
@@ -1,16 +1,20 @@
...
 from plexapi.base import Playable, PlexPartialObject, PlexSession
 from plexapi.exceptions import BadRequest
-from plexapi.mixins import (
-    RatingMixin,
-    ArtUrlMixin, ArtMixin, PosterUrlMixin, PosterMixin,
-    PhotoalbumEditMixins, PhotoEditMixins
-)
+from plexapi.mixins import (ArtMixin, ArtUrlMixin, PhotoalbumEditMixins,
+                            PhotoEditMixins, PosterMixin, PosterUrlMixin,
+                            RatingMixin)
```
JonnyWong16, line 14: Don't change in this PR.

plexapi/server.py
```diff
@@ -1,33 +1,40 @@
...
 from functools import cached_property
+from typing import TYPE_CHECKING, Optional, Union
 from urllib.parse import urlencode
 from xml.etree import ElementTree
 
 import requests
+from requests.status_codes import _codes as codes
 
-from plexapi import BASE_HEADERS, CONFIG, TIMEOUT, log, logfilter
+# Need these imports to populate utils.PLEXOBJECTS
```
JonnyWong16, line 14: Don't change in this PR. Move imports back to bottom.

JonnyWong16 (changes requested):

plexapi/base.py
Dr-Blank: `.query` can return `None` and `.findItems` does not accept `None`. so it should be handled in some way. returned empty list as `.fetchItems` returns a list

plexapi/base.py
Dr-Blank:
```py
def fetchItems(
        self,
        ekey: Union[str, List[int]],
        cls: Optional[Type[PlexObjectT]] = None,
        container_start: Optional[int] = None,
        container_size: Optional[int] = None,
        maxresults: Optional[int] = None,
        **kwargs: Any,
    ):
```

`ekey` is never `None` as defined by the method, hence unnecessary check. could typehint to accept `None` but then calling `fetchItems` without passing `ekey` does not make sense for the method, `None` should be handled by the caller imo

plexapi/base.py
Dr-Blank: as long as query could return `None`, handling of `None` should be done, it is done here as next step requires subscripting `data`

` self._loadData(data[0])`

plexapi/base.py
Dr-Blank:
```py
   for elem in data:  # None not iterable
```

plexapi/base.py
Dr-Blank: being assigned after few lines as below and not used before that assignment and this removal, hence the statement has no effect

```py
self._details_key = self._buildDetailsKey()
```

Dr-Blank:
@JonnyWong16 

can attr `key` be None in any case? if so, for which cases and is `key=None` useful? as I believe all media items must have a key, hence key cannot be `None` for all of those cases. correct me if I am wrong, but if that is the case, a base class for those items should make it `key` compulsory and for objects that have `key` as None should not inherit from this class with attr `key` in it.

I am saying this as it will save a ton of `if self.key is None` checks in methods which actually rely on `key`

plexapi/base.py
Dr-Blank: on a second thought maybe returning `result` is a better choice instead of empty list? or should raise error? what is the preferred way to handle no date? other choice is to make query raise an error instead of returning `None`

plexapi/base.py
glensc: eeh. why? any reference?

as it works in my project:
- https://github.com/Taxel/PlexTraktSync/blob/598685fe5fc4f3ae7acd7927218f5f9bc7a71352/setup.cfg#L20
- https://github.com/Taxel/PlexTraktSync/blob/598685fe5fc4f3ae7acd7927218f5f9bc7a71352/plextraktsync/util/execx.py#L6

pyupgrade config:
- https://github.com/Taxel/PlexTraktSync/blob/598685fe5fc4f3ae7acd7927218f5f9bc7a71352/.pre-commit-config.yaml#L44-L49

jjlawren: You need `from __future__ import annotations`.

glensc: I think what happens here is that if you import `__annotations__`, you can use `list[]`:
- https://stackoverflow.com/questions/39458193/using-list-tuple-etc-from-typing-vs-directly-referring-type-as-list-tuple-etc/62033243#62033243

unless it's also a generic:
- https://stackoverflow.com/questions/39458193/using-list-tuple-etc-from-typing-vs-directly-referring-type-as-list-tuple-etc/39458225#39458225

Dr-Blank:
should I break this PR further down? Maybe on a per function basis? That should make it easier for review as currently this change is maybe so huge that it is daunting to start reviewing, and multiple threads going on for each change.

Dr-Blank:
> I also created a new branch for `typehints`. I think it would be better to finish type hinting the entire library before merging into `master`.

I'm considering whether it would be more beneficial to merge it into `master` directly. Here are my thoughts:

1. Even adding type hints to only `base.py` could greatly enhance readability, providing immediate benefits to users and contributors with making new PRs.
1. By the time the entire project is type hinted, both branches may have diverged significantly, potentially leading to merge conflicts. Merging directly could mitigate this issue.
2. I'm planning to break down PRs on a per-function basis. Since type hints are primarily for developers and don't affect code execution, they're unlikely to introduce breaking changes.

@JonnyWong16, I'd appreciate your thoughts on this. Do you think the benefits of direct merging outweigh any potential risks?

either way I will close this PR and start anew since this is too huge an undertaking for review.

C0rn3j:
Why is this marked as a draft?

Even if it is incomplete, it should be merged, as it is useful on its own.

Thanks for your work, I was about to do this myself.

> I wonder if you ruled out providing a stub for typeshed?

While I am not OP - All the currently supported versions of Python support typing, I don't see the point of providing a stub.

Dr-Blank:
> Why is this marked as a draft?

I was new and made this a very big PR and hence I assume it was a lot of work to review. Hence I marked it draft as it was not moving ahead and I stopped using this library.

miaLDN:
>> I wonder if you ruled out providing a stub for typeshed?
> While I am not OP - All the currently supported versions of Python support typing, I don't see the point of providing a stub.

@C0rn3j rn

You mis-read (whoosh).  Any typing is better than none; inline typing the best; stub better than nothing.  Repeating what I already said and you already didn't understand.

PR #1350, opened 2024-01-30
Fix: Update guid filter examples for fetchItems

glensc:
## Description

Some why `imdb://` or `themoviedb://` regex does not filter the items, feels like pattern has "^" in it added by default.

I checked the code and it's not:
- https://github.com/pkkid/python-plexapi/blob/4780026a73af06af8801dada01f249ef6f8bb8db/plexapi/base.py#L29-L30

unless the `re.match` does it magically.
but using a complete agent name just works!

also changes `iregex`-> `regex`, the guids are always lowercase, no need for case insensitive match

additionally adds "tt12345" for legacy imdb entries

and `<Guid id="xxx">` tag matching example.

## Type of change

## Checklist:

### DIFF
--- a/plexapi/base.py
+++ b/plexapi/base.py
@@ -227,7 +227,7 @@ class PlexObject:
 
                     fetchItem(ekey, viewCount__gte=0)
                     fetchItem(ekey, Media__container__in=["mp4", "mkv"])
-                    fetchItem(ekey, guid__regex=r"(com\.plexapp\.agents\.imdb|themoviedb)://|tt\d+")
+                    fetchItem(ekey, guid__regex=r"com\.plexapp\.agents\.(imdb|themoviedb)://|tt\d+")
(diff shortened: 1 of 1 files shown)


JonnyWong16 (approved):

glensc:
> Some why imdb:// or themoviedb:// regex does not filter the items, feels like pattern has "^" in it added by default.

@JonnyWong16 can you comment why it behaves like that? i did not find code that would support that.

also thanks for the merge.

JonnyWong16:
`re.match()` matches the start of a string. 

https://docs.python.org/3/library/re.html#re.match

glensc:
So, using `re.match` rather `re.search` was a design error and fixing now is breaking change?

JonnyWong16:
I think doing `q.lower()` makes sense.

glensc:
q.lower fix:
- https://github.com/pkkid/python-plexapi/pull/1359

PR #1358, opened 2024-02-05
Breaking: Change regex/iregex to use re.search instead of re.match

glensc:
## Description

Change `fetchItem`'s `regex` and `iregex` filters to use `re.search` rather `re.match`.

This means that the pattern is now matched from anywhere the string, not start of the string:
- https://docs.python.org/3/library/re.html#search-vs-match

The user can specify `^` in their regex if they want to continue to match the start of the string.

See:
- https://github.com/pkkid/python-plexapi/pull/1350#issuecomment-1927772486

## Type of change

## Checklist:

### DIFF
--- a/plexapi/base.py
+++ b/plexapi/base.py
@@ -26,8 +26,8 @@ OPERATORS = {
     'endswith': lambda v, q: v.endswith(q),
     'iendswith': lambda v, q: v.lower().endswith(q),
     'exists': lambda v, q: v is not None if q else v is None,
-    'regex': lambda v, q: re.match(q, v),
-    'iregex': lambda v, q: re.match(q, v, flags=re.IGNORECASE),
+    'regex': lambda v, q: re.search(q, v),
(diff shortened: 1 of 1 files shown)


plexapi/base.py
```diff
@@ -26,8 +26,8 @@
     'endswith': lambda v, q: v.endswith(q),
     'iendswith': lambda v, q: v.lower().endswith(q),
     'exists': lambda v, q: v is not None if q else v is None,
-    'regex': lambda v, q: re.match(q, v),
-    'iregex': lambda v, q: re.match(q, v, flags=re.IGNORECASE),
+    'regex': lambda v, q: re.search(q, v),
+    'iregex': lambda v, q: re.search(q, v, flags=re.IGNORECASE),
```
JonnyWong16, line 30: Also cast to `bool`.
```suggestion
    'regex': lambda v, q: bool(re.search(q, v)),
    'iregex': lambda v, q: bool(re.search(q, v, flags=re.IGNORECASE)),
```

JonnyWong16 (approved):

PR #1361, opened 2024-02-11
feat: Add sonicAdventure method to MusicSection

### LINKED ISSUE #1360: feat: Endpoint for computing sonic adventure

api missing this endpoint `/library/sections/3/computePath?startID=106109&endID=97327&excludeElements=Media,Genre,Mood,Style&includeFields=thumbBlurHash`

_No response_

### Code Snippets

_No response_

### Additional Context

Notes for me
- need to mess around with this endpoint, the ids are of track for now. albums / artist could work
- work out what exclude elements does

Dr-Blank:
Fixes #1360

## Type of change

## Checklist:

### DIFF
--- a/plexapi/audio.py
+++ b/plexapi/audio.py
@@ -1,4 +1,6 @@
 # -*- coding: utf-8 -*-
+from __future__ import annotations
+
 import os
 from pathlib import Path
 from urllib.parse import quote_plus
@@ -526,6 +528,22 @@ class Track(
         """ Returns the Plex Media Server data directory where the metadata is stored. """
         guid_hash = utils.sha1hash(self.parentGuid)
         return str(Path('Metadata') / 'Albums' / guid_hash[0] / f'{guid_hash[1:]}.bundle')
+
+    def sonicAdventure(
+        self: TAudio,
+        to: TAudio,
+        **kwargs: Any,
+    ) -> list[TAudio]:
+        """Returns a sonic adventure from the current track to the specified track.
+
+        Parameters:
+            to: The target track for the sonic adventure.
+            **kwargs: Additional options passed into :func:`~plexapi.library.MusicSection.sonicAdventure`.
+
+        Returns:
+            List[:class:`~plexapi.audio.Track`]: list of tracks in the sonic adventure.
+        """
+        return self.section().sonicAdventure(self, to, **kwargs)
 
 
 @utils.registerPlexObject
--- a/plexapi/audio.py
+++ b/plexapi/audio.py
@@ -19,6 +19,7 @@ from plexapi.playlist import Playlist
 
 
 TAudio = TypeVar("TAudio", bound="Audio")
+TTrack = TypeVar("TTrack", bound="Track")
 
 
 class Audio(PlexPartialObject, PlayedUnplayedMixin):
@@ -530,14 +531,14 @@ class Track(
         return str(Path('Metadata') / 'Albums' / guid_hash[0] / f'{guid_hash[1:]}.bundle')
 
     def sonicAdventure(
-        self: TAudio,
-        to: TAudio,
+        self: TTrack,
+        to: TTrack,
         **kwargs: Any,
-    ) -> list[TAudio]:
+    ) -> list[TTrack]:
         """Returns a sonic adventure from the current track to the specified track.
 
         Parameters:
-            to: The target track for the sonic adventure.
+            to (:class:`~plexapi.audio.Track`): The target track for the sonic adventure.
             **kwargs: Additional options passed into :func:`~plexapi.library.MusicSection.sonicAdventure`.
 
         Returns:
--- a/plexapi/library.py
+++ b/plexapi/library.py
@@ -2047,16 +2047,16 @@ class MusicSection(LibrarySection, ArtistEditMixins, AlbumEditMixins, TrackEditM
             **kwargs: Any,
     ) -> list[Track]:
         """ Returns a list of tracks from this library section that are part of a sonic adventure.
-        ID's should be of a track, other ID's will return an empty list or items itself or an error.
+            ID's should be of a track, other ID's will return an empty list or items itself or an error.
 
-                Parameters:
-                    start (Track | int): The :class:`~plexapi.audio.Track` or ID of the first track in the sonic adventure.
-                    end (Track | int): The :class:`~plexapi.audio.Track` or ID of the last track in the sonic adventure.
-                    kwargs: Additional parameters to pass to :func:`~plexapi.base.PlexObject.fetchItems`.
+            Parameters:
+                start (Track | int): The :class:`~plexapi.audio.Track` or ID of the first track in the sonic adventure.
+                end (Track | int): The :class:`~plexapi.audio.Track` or ID of the last track in the sonic adventure.
+                kwargs: Additional parameters to pass to :func:`~plexapi.base.PlexObject.fetchItems`.
 
-                Returns:
-                    List[:class:`~plexapi.audio.Track`]: a list of tracks from this library section
-                    that are part of a sonic adventure.
+            Returns:
+                List[:class:`~plexapi.audio.Track`]: a list of tracks from this library section
+                that are part of a sonic adventure.
         """
         # can not use Track due to circular import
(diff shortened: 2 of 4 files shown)


Dr-Blank:
not sure if I understand why the tests failed for this change

plexapi/library.py
```diff
@@ -2033,6 +2040,26 @@ def sync(self, bitrate, limit=None, **kwargs):
         kwargs['policy'] = Policy.create(limit)
         return super(MusicSection, self).sync(**kwargs)
 
+    def sonicAdventure(
+            self,
+            startID: int,
+            endID: int,
+            **kwargs: Any,
+    ) -> list[Track]:
```
JonnyWong16, line 2048: I think `sonicAdventure()` should accept both `Track` objects and rating keys as arguments.

`sonicAdventure(start=track1, end=track2)`

Are `kwargs` necessary? What `kwargs` are expected?

JonnyWong16 (changes requested):
I believe sonic adventure is a Plex Pass feature so it will need to be an authenticated test.

It might also be useful to add a convenience method to tracks.

```python
def sonicAdventure(self, to):
    self.section().sonicAdventure(start=self, end=to)
```

```
>>> track1.sonicAdventure(to=track2)
```

plexapi/library.py
Dr-Blank: > Are kwargs necessary? What kwargs are expected?
any kwargs that could be passed to fetchitems, maybe a case where one needs to use filters to filter out tracks based on some param, like DJ stretch but with some other filtering mechanism.

Dr-Blank:
> I believe sonic adventure is a Plex Pass feature so it will need to be an authenticated test.
> 
> It might also be useful to add a convenience method to tracks.
> 
> ```python
> def sonicAdventure(self, to):
>     self.section().sonicAdventure(start=self, end=to)
> ```
> 
> ```
> >>> track1.sonicAdventure(to=track2)
> ```

I did not add this in this PR as to not clutter it like I have a habit to, which makes it more cumbersome to review. was holding it off for once this was merged/approved

plexapi/audio.py
```diff
@@ -527,6 +529,22 @@ def metadataDirectory(self):
...
+    def sonicAdventure(
+        self: TAudio,
+        to: TAudio,
+        **kwargs: Any,
+    ) -> list[TAudio]:
+        """Returns a sonic adventure from the current track to the specified track.
+
+        Parameters:
+            to: The target track for the sonic adventure.
+            **kwargs: Additional options passed into :func:`~plexapi.library.MusicSection.sonicAdventure`.
```
JonnyWong16, line 541: Add types to doc string.

tests/test_audio.py
```diff
@@ -398,6 +399,14 @@ def test_audio_Track_lyricStreams(track):
     assert not track.lyricStreams()
 
 
+@pytest.mark.authenticated
+def test_audio_Track_sonicAdventure(music):
```
JonnyWong16, line 403: Do these tests work on the bootstrap test server? It only has one track and we disable all the analysis tasks (although it looks like we forgot to disable sonic analysis).

https://github.com/pkkid/python-plexapi/blob/b51aa1924b5b28ba61aefbb4162aff6e6864561b/tools/plex-bootstraptest.py#L502-L503

Also needs a Plex Pass account.

```suggestion
def test_audio_Track_sonicAdventure(account_plexpass, music):
```

tests/test_library.py
```diff
@@ -430,6 +430,14 @@ def test_library_MusicSection_recentlyAdded(music, artist):
     assert track in music.recentlyAddedTracks()
 
 
+@pytest.mark.authenticated
+def test_library_MusicSection_sonicAdventure(music):
```
JonnyWong16, line 434:
```suggestion
def test_library_MusicSection_sonicAdventure(account_plexpass, music):
```

JonnyWong16 (changes requested):

plexapi/library.py
```diff
@@ -2033,6 +2040,31 @@ def sync(self, bitrate, limit=None, **kwargs):
...
+
+                Parameters:
+                    start (Track | int): The :class:`~plexapi.audio.Track` or ID of the first track in the sonic adventure.
+                    end (Track | int): The :class:`~plexapi.audio.Track` or ID of the last track in the sonic adventure.
+                    kwargs: Additional parameters to pass to :func:`~plexapi.base.PlexObject.fetchItems`.
+
+                Returns:
+                    List[:class:`~plexapi.audio.Track`]: a list of tracks from this library section
+                    that are part of a sonic adventure.
+        """
```
JonnyWong16, line 2060: Documentation is rendering incorrectly.

```suggestion
        """ Returns a list of tracks from this library section that are part of a sonic adventure.
            ID's should be of a track, other ID's will return an empty list or items itself or an error.

            Parameters:
                start (Track | int): The :class:`~plexapi.audio.Track` or ID of the first track in the sonic adventure.
                end (Track | int): The :class:`~plexapi.audio.Track` or ID of the last track in the sonic adventure.
                kwargs: Additional parameters to pass to :func:`~plexapi.base.PlexObject.fetchItems`.

            Returns:
                List[:class:`~plexapi.audio.Track`]: a list of tracks from this library section
                that are part of a sonic adventure.
        """
```

plexapi/audio.py
Dr-Blank: yes, upper bound should be Track but Audio works too since `self` will always be a Track and `to` will expect it to be of the same class

tests/test_audio.py
Dr-Blank: it might not, I changed my bootstrap server by adding a few tracks from my collection to test these features

JonnyWong16 (approved):

tests/test_audio.py
JonnyWong16: So this does not work on the bootstrap server with a single track, or even two tracks.

The tests fail at `assert len(adventure)` because there are no computed tracks. I think we just check if it is a list `assert isinstance(adventure, list)` instead.

plexapi/library.py
```diff
@@ -2033,6 +2040,31 @@ def sync(self, bitrate, limit=None, **kwargs):
...
+
+            Returns:
+                List[:class:`~plexapi.audio.Track`]: a list of tracks from this library section
+                that are part of a sonic adventure.
+        """
+        # can not use Track due to circular import
+        startID = start if isinstance(start, int) else start.ratingKey
+        endID = end if isinstance(end, int) else end.ratingKey
+
+        key = f"/library/sections/{self.key}/computePath?startID={startID}&endID={endID}"
```
JonnyWong16, line 2065: It will make it slightly quicker if you add `excludeElements=Media,Genre,Mood,Style`.

```suggestion
        params = {
            'startID': startID,
            'endID': endID,
            'excludeElements': 'Media,Genre,Mood,Style',
        }
        key = f"/library/sections/{self.key}/computePath{utils.joinArgs(params)}"
```

tests/test_audio.py
Dr-Blank: > So this does not work on the bootstrap server with a single track, or even two tracks.

would return list of 2 tracks, same as the args if not adventure between them.

JonnyWong16: I tried this on my own server and it returns a blank list. The start and end are not included.

Dr-Blank: okay, I would be wrong on this then, however the tests should be okay regardless

tests/test_library.py
```diff
@@ -430,6 +430,13 @@ def test_library_MusicSection_recentlyAdded(music, artist):
     assert track in music.recentlyAddedTracks()
 
 
+@pytest.mark.authenticated
+def test_library_MusicSection_sonicAdventure(account_plexpass, music):
+    tracks = music.searchTracks()
+    adventure = music.sonicAdventure(tracks[0], tracks[-1].ratingKey)
```
blacktwin, line 436: the first track needs to be using the `ratingKey` attribute here too.

`music.sonicAdventure(tracks[0].ratingKey, tracks[-1].ratingKey)`

JonnyWong16: We are testing that `sonicAdventure` accepts both `Track` objects and `int` rating keys.

PR #1373, opened 2024-02-21
`MediaContainer` class is now a list

### LINKED ISSUE #1372: `totalSize` of MediaContainer

api returns a bunch of metadata along with items fetched one of which is `totalSize`

it is useful in showing totalItems on the server without fetching all of them

other attributes are `offset` and `size`, standard pagination stuff

`fetchItems` already uses this info
https://github.com/pkkid/python-plexapi/blob/abcab4f78d22e86b380327c6014d3148e0248e6d/plexapi/base.py#L258

### Code Snippets

```python
tracks = music_section.searchTracks(maxresults=100)
print(f'showing {len(tracks)} of {tracks.totalSize} starting at {tracks.offset}')
```

### Additional Context

image
as on plex web

Dr-Blank:
## Description

`fetchItems` returns a MediaContainer, exactly how server does and is also a python list! Hence it would not break any current api.
additionally, it now supports totalSize attr

The only breaking change is that MediaContainer can now only accept `initpath` and `parent` as keword-only arguments in its initialisation, which I don't think anyone would be using.

Fixes #1372 
## Type of change

## Checklist:

### DIFF
--- a/plexapi/base.py
+++ b/plexapi/base.py
@@ -1069,6 +1069,22 @@ class MediaContainer(
                 self.offset if self.offset is not None else __iterable.offset
             )
 
+        # for all other attributes, overwrite with the new iterable's values if previously None
+        for key in (
+            "allowSync",
+            "augmentationKey",
+            "identifier",
+            "librarySectionID",
+            "librarySectionTitle",
+            "librarySectionUUID",
+            "mediaTagPrefix",
+            "mediaTagVersion",
+        ):
+            if (not hasattr(self, key)) or (getattr(self, key) is None):
+                if not hasattr(__iterable, key):
+                    continue
+                setattr(self, key, getattr(__iterable, key))
+
     def _loadData(self, data):
         self._data = data
         self.allowSync = utils.cast(int, data.attrib.get('allowSync'))
--- a/plexapi/base.py
+++ b/plexapi/base.py
@@ -1,12 +1,20 @@
 # -*- coding: utf-8 -*-
 import re
+from typing import TYPE_CHECKING, Generic, Iterable, List, Optional, TypeVar, Union
 import weakref
 from functools import cached_property
 from urllib.parse import urlencode
 from xml.etree import ElementTree
+from xml.etree.ElementTree import Element
 
 from plexapi import CONFIG, X_PLEX_CONTAINER_SIZE, log, utils
 from plexapi.exceptions import BadRequest, NotFound, UnknownType, Unsupported
+
+if TYPE_CHECKING:
+    from plexapi.server import PlexServer
+
+PlexObjectT = TypeVar("PlexObjectT", bound='PlexObject')
+MediaContainerT = TypeVar("MediaContainerT", bound="MediaContainer")
 
 USER_DONT_RELOAD_FOR_KEYS = set()
 _DONT_RELOAD_FOR_KEYS = {'key'}
@@ -245,8 +253,7 @@ class PlexObject:
         if maxresults is not None:
             container_size = min(container_size, maxresults)
 
-        results = []
-        subresults = []
+        results = MediaContainer[cls](self._server, Element('MediaContainer'), initpath=ekey)
         headers = {}
 
         while True:
@@ -324,7 +331,7 @@ class PlexObject:
         if rtag:
             data = next(utils.iterXMLBFS(data, rtag), [])
         # loop through all data elements to find matches
-        items = []
+        items = MediaContainer[cls](self._server, data, initpath=initpath) if data.tag == 'MediaContainer' else []
         for elem in data:
             if self._checkAttrs(elem, **kwargs):
                 item = self._buildItemOrNone(elem, cls, initpath)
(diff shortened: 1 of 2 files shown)


Dr-Blank:
could you guide me as to where to put the test for this?
also how would the test be implemented?
for example test fetchItems directly
or indirectly, like searchepisodes, and test on the return

tests i can think of:
 1. isinstance list
 2. size, offset, totalSize are int

plexapi/base.py
```diff
@@ -1009,11 +1020,55 @@ class MediaContainer(PlexObject):
...
 
+    def __init__(
+        self,
+        server: "PlexServer",
+        data: Element,
+        *args: PlexObjectT,
+        initpath: Optional[str] = None,
+        parent: Optional[PlexObject] = None,
+    ) -> None:
+        # super calls Generic.__init__ which calls list.__init__ eventually
```
Dr-Blank, line 1038: `mro` suggests otherwise, please review

plexapi/base.py
```diff
@@ -1009,11 +1020,55 @@ class MediaContainer(PlexObject):
...
+        # update size, totalSize, and offset
+        if not isinstance(__iterable, MediaContainer):
+            return
+
+        # prefer the totalSize of the new iterable even if it is smaller
+        self.totalSize = (
+            __iterable.totalSize
+            if __iterable.totalSize is not None
+            else self.totalSize
+        )  # ideally both should be equal
```
Dr-Blank, line 1057: newest value will be the latest from the server, but total size should not change in such short time

plexapi/base.py
```diff
@@ -1009,11 +1020,55 @@ class MediaContainer(PlexObject):
...
+
+        # the size of the new iterable is added to the current size
+        self.size = curr_size + (
```
Dr-Blank, line 1060: totalsize get increased as more and more items are fetched

plexapi/base.py
```diff
@@ -1009,11 +1020,55 @@ class MediaContainer(PlexObject):
...
+            __iterable.size if __iterable.size is not None else len(__iterable)
+        )
+
+        # the offset is the minimum of the two, prefering older values
+        if self.offset is not None and __iterable.offset is not None:
+            self.offset = min(self.offset, __iterable.offset)
+        else:
+            self.offset = (
+                self.offset if self.offset is not None else __iterable.offset
+            )
```
Dr-Blank, line 1070: if we start fetching from 0 and fetch 50 items, and next fetch offset is 50, however when etending, the offset is still from 0, but now the size is 50 + 50 =100

Dr-Blank:
I suppose regression tests are not needed for this PR since the current tests should be enough.

I have added some tests pertaining to new functionality only.

JonnyWong16:
This looks greats. It seems to be mostly working, but is it possible to load the other `MediaContainer` attributes as well?

```py
>> results = plex.library.section("Movies (Test)").all()  # results is now a MediaContainer
>> results.librarySectionID  # this is None
```

```xml
<MediaContainer size="17" allowSync="1" art="/:/resources/movie-fanart.jpg" identifier="com.plexapp.plugins.library" librarySectionID="41" librarySectionTitle="Movies (Test)" librarySectionUUID="e9535986-3fd3-4f88-9793-77a50b04dd01" mediaTagPrefix="/system/bundle/media/flags/" mediaTagVersion="1708667686" thumb="/:/resources/movie.png" title1="Movies (Test)" title2="All Movies (Test)" viewGroup="movie" viewMode="65592">
```

`results` is initialized with empty `data = Element('MediaContainer')` which never gets updated with the attributes from the XML.
`subresults` gets initialized with the actual data from the XML (from `items` inside `findItems()`).

Dr-Blank:
> is it possible to load the other `MediaContainer` attributes as well?

https://github.com/pkkid/python-plexapi/pull/1373/commits/06819fa614ce3d9d84f0f51980094cc43569b22e solves this

JonnyWong16 (approved):

plexapi/base.py
```diff
@@ -324,7 +331,7 @@ def findItems(self, data, cls=None, initpath=None, rtag=None, **kwargs):
         if rtag:
             data = next(utils.iterXMLBFS(data, rtag), [])
         # loop through all data elements to find matches
-        items = []
+        items = MediaContainer[cls](self._server, data, initpath=initpath) if data.tag == 'MediaContainer' else []
```
JonnyWong16, line 334: `AttributeError: 'list' object has no attribute 'tag'` when `data = []` from the line above.

Dr-Blank: Maybe we can initialise an empty media container when the tag attribute does not exist? I'll create a pr soon if not already done.

JonnyWong16: Initialize an `empty` element like what you did here:

https://github.com/pkkid/python-plexapi/blob/d6754c0cdeec67056df10318e708ca6ec4a61bb0/plexapi/base.py#L459

Dr-Blank: should be resolved in #1396

Issue #1375, opened 2024-02-24
createPlaylist does not guard smart=True and non-empty items


glensc:
### Describe the Bug

Seems `server.createPlaylist(self.name, items=items, smart=True, section=section)` does not guard `smart=True` and non empty `items`.

While other methods like `playlist.removeItems()` and `playlist.addItems()` guard that you do not add items with smart playlist.

The outcome is that created smart playlist doesn't actually contain the items, but something more (haven't figured out what yet, seems like the whole library is added to such smart playlist)

See details of this exploration:
- https://github.com/Taxel/PlexTraktSync/pull/1832#issuecomment-1962421451

### Operating System and Version

Irrelevant

### Plex Media Server Version

Irrelevant

### Python Version

Irrelevant

### PlexAPI Version

4.15.10

JonnyWong16:
Thanks, can you submit a fix?

glensc:
https://github.com/pkkid/python-plexapi/pull/1377

PR #1377, opened 2024-02-26
Fix: Raise an exception when trying to create a smart collection/playlist with items

glensc:
## Description

Fixes https://github.com/pkkid/python-plexapi/issues/1375

## Type of change

## Checklist:

### DIFF
--- a/plexapi/playlist.py
+++ b/plexapi/playlist.py
@@ -434,6 +434,8 @@ class Playlist(
         if m3ufilepath:
             return cls._createFromM3U(server, title, section, m3ufilepath)
         elif smart:
+            if items:
+                raise BadRequest('Cannot fill items for a smart playlist.')
             return cls._createSmart(server, title, section, limit, libtype, sort, filters, **kwargs)
         else:
             return cls._create(server, title, items)
--- a/plexapi/playlist.py
+++ b/plexapi/playlist.py
@@ -435,7 +435,7 @@ class Playlist(
             return cls._createFromM3U(server, title, section, m3ufilepath)
         elif smart:
             if items:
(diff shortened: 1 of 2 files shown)


glensc:
actually, i'm not sure what was happening. as the "items" doesn't seem to be passed to the `_createSmart` method

glensc:
so, what was happening was that it called this URL:

- `/playlists?smart=1&title=The Melchior Collection&type=video&uri=server://[redacted]/com.plexapp.plugins.library/library/sections/1/all?includeGuids=1&type=1`


which seems like smart playlist with no filters (so it contains whole library)

glensc:
Maybe what could be done here instead is to raise an error if no filters are provided

glensc:
So the change here should be fine. It would disallow using smart and items, possibly not confusing the user. 

user can still create smart playlist without filters as long as they don't also pass items.

plexapi/playlist.py
```diff
@@ -434,6 +434,8 @@ def create(cls, server, title, section=None, items=None, smart=False, limit=None
         if m3ufilepath:
             return cls._createFromM3U(server, title, section, m3ufilepath)
         elif smart:
+            if items:
+                raise BadRequest('Cannot fill items for a smart playlist.')
```
JonnyWong16, line 438:
```suggestion
                raise BadRequest('Cannot create a smart playlist with items.')
```

JonnyWong16 (changes requested):
Could you also add the same exception for smart collections?

https://github.com/pkkid/python-plexapi/blob/abcab4f78d22e86b380327c6014d3148e0248e6d/plexapi/collection.py#L505-L506

JonnyWong16 (approved):

PR #1394, opened 2024-04-09
Fix: stop maxresults from overwriting previous results #1393

### LINKED ISSUE #1393: Using `maxresults` in fetchItems() tramples over results
### Describe the Bug

`plexapi.server.library.fetchItems()` supports a `maxresults` parameter which, when specified, limits the number of results returned. In code, this is implemented by reducing the size of the container used when there are fewer than `container_size` results that need to be fetched.

If `maxresults` were set to 123, for example, then the first fetch would get 100 items (because of the default container size), and the second only 23. In this example, on the second fetch the container size would be reduced to 23.

At the end of the `fetchItems` loop, `container_start` is advanced by `container_size` ready to receive the next batch of results. Unfortunately this is done *after* `container_size` is modified to support `maxresults`. So, in the example above, `container_start` is set to 23, so the next batch of results overwrites whatever results were already 'at 23' in the container.

### Code Snippets

```python
# where plex is an authorised instance of plexapi.server and Plex has more than 110 TV shows:

def demonstrate_duplicates(self) -> None:
    def build_shows_by_id(shows: list) -> dict:
        shows_by_id = {}
        for show in shows:
            if show.key in shows_by_id:
                print(f"Duplicate: {show.key} = {show.title}")
            else:
                shows_by_id[show.key] = show
        return shows_by_id

    items = plex.library.search(libtype="show", limit=110)
    ids = [item.key for item in items]

    int_ids = [int(id.rsplit("/", 1)[-1]) for id in ids]
    bad_shows = plex.library.fetchItems(ekey=int_ids, maxresults=len(int_ids))
    good_shows = plex.library.fetchItems(ekey=int_ids)

    good_shows_by_id = build_shows_by_id(good_shows)
    bad_shows_by_id = build_shows_by_id(bad_shows)
```

### Expected Behavior

`fetchItems()` in PlexObject would return the media requested whether `maxresults` is specified or not.

### Additional Context

_No response_

### Operating System and Version

macOS 14.3.1

### Plex Media Server Version

4.125.1

### Python Version

3.12.2

### PlexAPI Version

4.15.11

Touchstone64:
## Description

Please include a summary of the change and which issue is fixed.

Fixes #1393

## Type of change

## Checklist:

### DIFF
--- a/plexapi/base.py
+++ b/plexapi/base.py
@@ -283,6 +283,8 @@ class PlexObject:
 
             results.extend(subresults)
 
+            last_container_size = container_size
+
             wanted_number_of_items = total_size - offset
             if maxresults is not None:
                 wanted_number_of_items = min(maxresults, wanted_number_of_items)
@@ -291,7 +293,7 @@ class PlexObject:
             if wanted_number_of_items <= len(results):
                 break
 
-            container_start += container_size
+            container_start += last_container_size
 
             if container_start > total_size:
                 break
--- a/plexapi/base.py
+++ b/plexapi/base.py
@@ -283,7 +283,10 @@ class PlexObject:
 
             results.extend(subresults)
 
-            last_container_size = container_size
+            container_start += container_size
+
+            if container_start > total_size:
+                break
 
             wanted_number_of_items = total_size - offset
             if maxresults is not None:
@@ -291,11 +294,6 @@ class PlexObject:
                 container_size = min(container_size, wanted_number_of_items - len(results))
 
             if wanted_number_of_items <= len(results):
-                break
-
-            container_start += last_container_size
-
-            if container_start > total_size:
                 break
 
         return results
--- a/tests/test_library.py
+++ b/tests/test_library.py
@@ -127,12 +127,12 @@ def test_library_fetchItems_with_maxresults(plex):
     ids = [item.key for item in items1]
     # Reduce '/library/metadata/123' to '123'
     int_ids = [int(id.rsplit("/", 1)[-1]) for id in ids]
-    items2 = [item.key for item in plex.library.fetchItems(container_size=size, ekey=int_ids)]
-    items3 = [item.key for item in plex.library.fetchItems(
+    keys1 = [item.key for item in plex.library.fetchItems(container_size=size, ekey=int_ids)]
+    keys2 = [item.key for item in plex.library.fetchItems(
         container_size=size, ekey=int_ids, maxresults=len(int_ids)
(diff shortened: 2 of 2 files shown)


Touchstone64:
I'll look into adding a test for this 👍

plexapi/base.py
```diff
@@ -283,6 +283,8 @@ def fetchItems(self, ekey, cls=None, container_start=None, container_size=None,
 
             results.extend(subresults)
 
+            last_container_size = container_size
```
JonnyWong16, line 286: Isn't it okay to just move `container_start += container_size` up here before the `maxresults` logic below?

Touchstone64:
Unit test added for #1394

plexapi/base.py
Touchstone64: Yep @JonnyWong16 that was my first thought too. Then I thought, what if later calculations based on container_size needed to be considered, but I'm probably over-worrying.

JonnyWong16: I think just keep it simple and move it above.

Touchstone64: Agreed, changes committed. New test passes, please excuse poor internet connection and consequently slow responses.

tests/test_library.py
```diff
@@ -120,6 +120,21 @@ def test_library_fetchItem(plex, movie):
...
+    size = len(items1) - 5
+    ids = [item.key for item in items1]
+    # Reduce '/library/metadata/123' to '123'
+    int_ids = [int(id.rsplit("/", 1)[-1]) for id in ids]
+    keys1 = [item.key for item in plex.library.fetchItems(container_size=size, ekey=int_ids)]
+    keys2 = [item.key for item in plex.library.fetchItems(
+        container_size=size, ekey=int_ids, maxresults=len(int_ids)
+    )]
+    assert len(keys1) == len(set(keys1))
+    assert len(keys2) == len(set(keys2))
```
JonnyWong16, line 135:
```suggestion
def test_library_fetchItems_with_maxresults(plex, tvshows):
    items = tvshows.searchEpisodes()
    assert len(items) > 5
    size = len(items) - 5
    ratingKeys = [item.ratingKey for item in items]
    items1 = plex.fetchItems(ekey=ratingKeys, container_size=size)
    items2 = plex.fetchItems(ekey=ratingKeys, container_size=size, maxresults=len(items))
    assert items1 == items2 == items
```

JonnyWong16 (changes requested):

PR #1437, opened 2024-07-02
add searchProviders parameter to searchDiscover method

mdinicola:
I found the searchDiscover method is broken.  Plex now requires a new parameter called searchProviders to be provided in the query.  This parameter is a comma separated string with the following values:

- discovery: This is the default behavior matching the previous functionality
- PLEXAVOD: Plex ad-supported video service
- PLEXTVOD: Plex transactional video service.  I've also seen this referred to as Plex Rentals.

I've updated the searchDiscover method to default to just "discovery", with the option for users to include the other providers if they wish.

### DIFF
--- a/plexapi/myplex.py
+++ b/plexapi/myplex.py
@@ -1057,7 +1057,7 @@ class MyPlexAccount(PlexObject):
                 query (str): Search query.
                 limit (int, optional): Limit to the specified number of results. Default 30.
                 libtype (str, optional): 'movie' or 'show' to only return movies or shows, otherwise return all items.
-                searchprovider (str, optional): 'discover' for default behavior 
+                searchprovider (str, optional): 'discover' for default behavior
                     or 'discover,PLEXAVOD' to also include the Plex ad-suported video service
                     or 'discover,PLEXAVOD,PLEXTVOD' to also include the Plex video rental service
         """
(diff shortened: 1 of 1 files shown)


JonnyWong16:
Please fix `flake8` trailing whitespace.

plexapi/myplex.py
```diff
@@ -1049,14 +1049,17 @@ def markUnplayed(self, item):
...
-    def searchDiscover(self, query, limit=30, libtype=None):
+    def searchDiscover(self, query, limit=30, libtype=None, searchProviders='discover'):
         """ Search for movies and TV shows in Discover.
             Returns a list of :class:`~plexapi.video.Movie` and :class:`~plexapi.video.Show` objects.
 
             Parameters:
                 query (str): Search query.
                 limit (int, optional): Limit to the specified number of results. Default 30.
                 libtype (str, optional): 'movie' or 'show' to only return movies or shows, otherwise return all items.
+                searchprovider (str, optional): 'discover' for default behavior
```
JonnyWong16, line 1060:
```suggestion
                searchProviders (str, optional): 'discover' for default behavior
```

JonnyWong16: Maybe even just name the argument `providers`.

JonnyWong16 (approved):

glensc (approved):

PR #1471, opened 2024-10-14
replace http client Requests to compatible Niquests

Ousret:
## Description

This PR effectively replace the http client Requests for [Niquests](https://github.com/jawah/niquests).
Niquests is a drop-in replacement for Requests that is no longer under feature freeze.

This new client support HTTP/2, and HTTP/3 by default and offers both sync and async interfaces. It supports all the latest shiny features you would expect from an http client.

If you were interested on merging, I will be thrilled to propose a followup PR that will propose a script that generate
the async part of plex-client automatically. 

This PR is a low footprint change. I did not change the "alert" part, based on WS, because there's something we can't support: "reusing existing socket instance". Would it be a breaking change according to you?

*disclaimer:* _I maintain Niquests._

## Type of change

## Checklist:

### DIFF
--- a/plexapi/client.py
+++ b/plexapi/client.py
@@ -3,13 +3,13 @@ import time
 import weakref
 from xml.etree import ElementTree
 
-import requests
+import niquests as requests
 
 from plexapi import BASE_HEADERS, CONFIG, TIMEOUT, log, logfilter, utils
 from plexapi.base import PlexObject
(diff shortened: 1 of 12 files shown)


glensc:
The http2/http3 is not use for plex media servers (support only http11), but async functions could have improvement to parallelism. can't comment plex.tv requests, haven't checked what they support.

PR #1484, opened 2024-11-17
Use SPDX license identifier

cdce8p:
## Description
Use the SPDX license identifier as recommended by [PEP 639](https://peps.python.org/pep-0639/).
https://spdx.org/licenses/BSD-3-Clause.html

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -7,7 +7,7 @@ description = "Python bindings for the Plex API."
 readme = "README.rst"
 requires-python = ">=3.9"
 keywords = ["plex", "api"]
-license = {file = "LICENSE.txt"}
+license = {text = "BSD-3-Clause"}
 classifiers = [
     "Operating System :: OS Independent",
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -7,7 +7,7 @@ description = "Python bindings for the Plex API."
 readme = "README.rst"
 requires-python = ">=3.9"
 keywords = ["plex", "api"]
-license = {file = "LICENSE.txt"}
+license = {text = "BSD-3-Clause"}
```
JonnyWong16, line 10: Should this just be

```suggestion
license = "BSD-3-Clause"
```

cdce8p: That's not yet supported by setuptools unfortunately. Using the legacy `license.text` is the next best thing for the time being.

_There is a open PR in setuptools to add full support for it. That one depends on some other (unrelated) work though, so it's still some time until it's added._

PR #1491, opened 2024-12-27
Adding History object and history deletion

blacktwin:
## Description

When requesting history the items object would contain a lot of unnecessary attributes. Adding the History object to only use the available attributes from the XML response. I discovered the Plex Dash allows for deleting history so I've added that functionally as well. 

To wipe the history from Library 1:
```python
history = server.history(librarySectionID=1)
for item in history:
    item.delete()
```

Testing is not applicable due to somewhat recent changes to creating history on a Plex server. 

## Type of change

## Checklist:

### DIFF
--- a/plexapi/server.py
+++ b/plexapi/server.py
@@ -676,7 +676,7 @@ class PlexServer(PlexObject):
             args['viewedAt>'] = int(mindate.timestamp())
 
         key = f'/status/sessions/history/all{utils.joinArgs(args)}'
-        return self.fetchItems(key, maxresults=maxresults)
+        return self.fetchItems(key, maxresults=maxresults, cls=History)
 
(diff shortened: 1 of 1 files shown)


JonnyWong16:
History should be using the `PlexHistory` object (and the subclasses `TrackHistory`, `MovieHistory`, and `EpisodeHistory`). Maybe there's a bug where it's not automatically using those objects. 

https://github.com/pkkid/python-plexapi/blob/f2310ea3d3416f0d34811abed47f6c87d9a90195/plexapi/base.py#L1013

Also side note, I was planning `accountID` as a parameter into `LibrarySection.history()` so you can get library history for someone other than the admin account.

blacktwin:
No bug. Removing my code and it works as expected. I just didn't look in the correct library👎. Closing.

JonnyWong16:
Is it still worth it to remove the extra attributes?

Issue #1509, opened 2025-04-03
Ability to load attributes on demand then to cache them instead of loading them on object instantiation


eliasbenb:
### What is your feature request?

After profiling my program that relies on python-plexapi, I found that `_loadData` was the function with the highest CPU time (~65% of the program's total CPU time).

This is largely because python-plexapi loads all available attributes on object instantiation. Most of which never get used. Instead, I suggest attributes get loaded on demand and then cached (with the cache key changing whenever the `_data` attribute changes).

For my use cases, where I only need access to the `ratingKey, guid, viewCount, lastViewedAt, rating` attributes. Using on demand access would cut down `_loadData`'s runtime by over 80% in this scenario (according to my rough estimate from the flame graph below).

**Pyspy profiler flame graph:**

![Image](https://github.com/user-attachments/assets/4003c53e-ef9c-4b6b-a149-5207a4613f07)

PR #1510, opened 2025-04-04
Lazy loading and caching for attributes set in `_loadData(..)`

eliasbenb:
## Description

This PR introduces lazy loading for data attributes in all Plex objects as well as a caching mechanism when loading those attributes using a new `@cached_data_property` decorator. This change significantly improves performance by deferring the loading of computationally expensive attributes until when the user requests it.

The cache of properties created with `@cached_data_property` will also automatically be invalidated when `_loadData(...)` is called with a new XML data object.

Fixes: #1509 

## Type of Change

## Checklist:

## Additional Notes

The initial performance testing is very promising.  Below is a test script that was ran against the production python-plexapi package and this PR's version:

```py
import timeit

def test():

    plex = PlexServer()

    res = []
    for section in plex.library.sections():
        items = section.all()
        res.extend([(i.title, i.lastViewedAt) for i in items])

if __name__ == "__main__":
    print(timeit.timeit(test, number=1))
```

In this testing scenario, the program does not need access to any of the expensive `Video` attributes (e.g. `images`, `genres`, `directors`). Since those expensive queries are skipped, the lazy-loaded testing results are ~3x faster than the production python-plexapi package.

```shell
# Time in seconds
Lazy-loaded: 5.732901584000501
Production: 17.07976305699958
```

Below are the Pyspy profiler outputs between the two packages against the above script:

**Lazy-loaded:**

lazy-loaded

**Production:**

production

As can clearly be seen, python-plexapi (purple) is no longer the bottleneck, instead, the requests and urllib packages (which are making the HTTP requests to the Plex server) are the new bottlenecks.

### DIFF
--- a/plexapi/base.py
+++ b/plexapi/base.py
@@ -532,14 +532,6 @@ class PlexObject(metaclass=PlexObjectMeta):
             return float(value)
         return value
 
-    def _invalidateCachedProperties(self):
-        """Invalidate all cached data property values."""
-        cached_props = getattr(self.__class__, '_cached_data_properties', set())
-
-        for prop_name in cached_props:
-            if prop_name in self.__dict__:
-                del self.__dict__[prop_name]
-
     def _invalidateCacheAndLoadData(self, data):
         """Load attribute values from Plex XML response and invalidate cached properties."""
         old_data_id = id(getattr(self, '_data', None))
@@ -550,6 +542,14 @@ class PlexObject(metaclass=PlexObjectMeta):
             self._invalidateCachedProperties()
 
         self._loadData(data)
+
+    def _invalidateCachedProperties(self):
+        """Invalidate all cached data property values."""
+        cached_props = getattr(self.__class__, '_cached_data_properties', set())
+
+        for prop_name in cached_props:
+            if prop_name in self.__dict__:
+                del self.__dict__[prop_name]
 
     def _loadData(self, data):
         """ Load attribute values from Plex XML response. """
--- a/plexapi/base.py
+++ b/plexapi/base.py
@@ -809,9 +809,11 @@ class PlexPartialObject(PlexObject):
 
 
 class Playable:
-    """ This is a general place to store functions specific to media that is Playable.
+    """ This is a mixin to store functions specific to media that is Playable.
         Things were getting mixed up a bit when dealing with Shows, Season, Artists,
         Albums which are all not playable.
+
+        This class 
 
         Attributes:
             playlistItemID (int): Playlist item ID (only populated for :class:`~plexapi.playlist.Playlist` items).
@@ -988,7 +990,7 @@ class Playable:
 
 
 class PlexSession:
-    """ This is a general place to store functions specific to media that is a Plex Session.
+    """ This is a mixin to store functions specific to media that is a Plex Session.
 
         Attributes:
             live (bool): True if this is a live tv session.
@@ -1091,7 +1093,7 @@ class PlexSession:
 
 
 class PlexHistory:
-    """ This is a general place to store functions specific to media that is a Plex history item.
+    """ This is a mixin to store functions specific to media that is a Plex history item.
 
         Attributes:
             accountID (int): The associated :class:`~plexapi.server.SystemAccount` ID.
--- a/plexapi/base.py
+++ b/plexapi/base.py
@@ -813,8 +813,6 @@ class Playable:
         Things were getting mixed up a bit when dealing with Shows, Season, Artists,
         Albums which are all not playable.
 
-        This class 
-
         Attributes:
             playlistItemID (int): Playlist item ID (only populated for :class:`~plexapi.playlist.Playlist` items).
             playQueueItemID (int): PlayQueue item ID (only populated for :class:`~plexapi.playlist.PlayQueue` items).
@@ -1005,10 +1003,7 @@ class PlexSession:
     def _loadData(self, data):
         """ Load attribute values from Plex XML response. """
         self.live = utils.cast(bool, data.attrib.get('live', '0'))
-        self.player = self.findItem(data, etag='Player')
-        self.session = self.findItem(data, etag='Session')
         self.sessionKey = utils.cast(int, data.attrib.get('sessionKey'))
-        self.transcodeSession = self.findItem(data, etag='TranscodeSession')
 
         user = data.find('User')
         self._username = user.attrib.get('title')
--- a/plexapi/base.py
+++ b/plexapi/base.py
@@ -1174,7 +1174,7 @@ class MediaContainer(
                 setattr(self, key, getattr(__iterable, key))
 
     def _loadData(self, data):
-        self._data = data
+        PlexPartialObject._loadData(self, data)
         self.allowSync = utils.cast(int, data.attrib.get('allowSync'))
         self.augmentationKey = data.attrib.get('augmentationKey')
         self.identifier = data.attrib.get('identifier')
--- a/plexapi/base.py
+++ b/plexapi/base.py
@@ -537,7 +537,7 @@ class PlexObject(metaclass=PlexObjectMeta):
         cached_props = getattr(self.__class__, '_cached_data_properties', set())
 
         for prop_name in cached_props:
-            cache_name = f"_{prop_name}"
+            cache_name = prop_name
             if cache_name in self.__dict__:
                 del self.__dict__[cache_name]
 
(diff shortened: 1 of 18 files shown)


JonnyWong16:
Yeah, I was just thinking about it when I saw your graph. A separate PR would be good.

JonnyWong16:
The cache invalidation isn't working.

```py

plex = PlexServer()
movie = plex.library.section("Movies").get("GoldenEye")

id(movie._data)  # 2701410045600
movie.__dict__['genres']  # KeyError as expected
movie.genres  # retrieved - [<Genre:Action>, <Genre:Adventure>]
movie.__dict__['genres']  # cached - [<Genre:Action>, <Genre:Adventure>]

movie.reload()
id(movie._data)  # 2701409346640 - ID changed
movie.__dict__['genres']  # still cached - [<Genre:Action>, <Genre:Adventure>]
```

plexapi/base.py
```diff
@@ -497,8 +532,23 @@ def _castAttrValue(self, op, query, value):
...
         return value
 
+    def _invalidateCachedProperties(self):
+        """Invalidate all cached data property values."""
+        cached_props = getattr(self.__class__, '_cached_data_properties', set())
+
+        for prop_name in cached_props:
+            cache_name = prop_name
+            if cache_name in self.__dict__:
+                del self.__dict__[cache_name]
```
JonnyWong16, line 542:
```suggestion
        for prop_name in cached_props:
            if prop_name in self.__dict__:
                del self.__dict__[prop_name]
```

JonnyWong16:
Still a bunch of tests failing.

JonnyWong16:
### ❓ `test_library_and_section_search_for_movie`:

This is only failing because the previous test for editing library locations failed so it ended up with two movies.

### 🚫 `test_photo_Photoalbum_mixins_fields`, `test_photo_Photo_mixins_fields`, `test_photo_Photo_mixins_tags`, `test_Playlist_mixins_fields`, `test_video_Season_mixins_tags`:

`fields` are a list of locked edits.

### ✅ `test_video_Movie_reload_kwargs`:

The reason I used `movie.__dict__['media']` here is because I explicitly don't want to trigger the auto-reload on accessing empty attributes via `movie.media`.

JonnyWong16:
Yes, that should work for the movie reload test.

JonnyWong16:
### 🚫 `test_library_add_edit_delete`:

* `LibrarySection` has it's own `reload()` method that needs to handle the caching invalidating separately.

### 🚫 `test_photo_Photoalbum_mixins_fields`, `test_photo_Photo_mixins_fields`, `test_photo_Photo_mixins_tags`, `test_Playlist_mixins_fields`, `test_video_Season_mixins_tags`:

* `Photoalbum`, `Photo`, and `Playlist` are missing the call to `PlexObject._loadData(self, data)`
* `Season` was fixed with 526e47b01e986d6a6a779246c5d341f5a1403c65

JonnyWong16:
You can add the two cache invalidation tests that you used above.

plexapi/media.py
```diff
@@ -1248,14 +1252,15 @@ def __repr__(self):
...
-        self._data = data
+        PlexObject._loadData(self, data)
         self.hasAttribution = data.attrib.get('hasAttribution')
         self.hasPrefs = data.attrib.get('hasPrefs')
         self.identifier = data.attrib.get('identifier')
         self.name = data.attrib.get('name')
         self.primary = data.attrib.get('primary')
         self.shortIdentifier = self.identifier.rsplit('.', 1)[1]
 
+        # TODO: How should the cached data property be handled here?
```
JonnyWong16, line 1263: Does this work?
```py
@cached_data_property
def languageCodes(self):
    if 'mediaType' in self._initpath:
        return self.listAttrs(data, 'code', etag='Language')
    return []

@cached_data_property
def mediaTypes(self):
    if 'mediaType' not in self._initpath:
        return self.findItems(data, cls=AgentMediaType)
    return []
```

JonnyWong16 (changes requested):
Please simplify the variable in the cache invalidation loop (see my comment above).

`base.PlexSession` has a few attributes that use `findItems` that can be turned into `cached_data_property`.

Since you are updating `_loadData`, could you also:
* Update the missing docstrings for all `_loadData`? Just adding `""" Load attribute values from Plex XML response. """`.
* Fix the use of `super(...)._loadData(data)` for `media.VideoStream`, `media.AudioStream`, `media.SubtitleStream`, `media.LyricStream`, and `video.Extra`

JonnyWong16:
> It looks like PlexSession doesn't inherit the PlexObject class, so the cached data property invalidation logic won't work without either 1) making it a PlexObject subclass, or 2) assigning it the PlexObjectMeta meta class. Either will work the same, I was just wondering if there's a reason it isn't a subclass of PlexObject already.

I think `Playable`, `PlexSession` and `PlexHistory` are treated more like a mixin than base classes. The only time `PlexSession` is ever inherited is in something that also inherits from `PlexObject`.

https://github.com/pushingkarmaorg/python-plexapi/blob/a8d051f29bbadca28dc9bd4bf50f1fb76d0fb77a/plexapi/video.py#L1204-L1213

I don't know if this will cause issues if both `Movie._loadData(self, data)` and `PlexSession._loadData(self, data)` handle the caching. And similarly with `Playable` and `PlexHistory`.

JonnyWong16:
> There's also another issue of whether _loadData(...) should do the self._data assignment. It makes sense aesthetically but also adds a lot of duplicated lines.
> 
> Putting it in _invalidateCacheAndLoadData(...) in simpler but possibly less clear.
> 
> I'm leaning towards leaving it in just _invalidateCacheAndLoadData(...) but what do you think?

I think it's fine. You need to get the `id()` of the old data before you reassign it.

JonnyWong16:
`flake8` tests are failing due to unused imports.

plexapi/base.py
```diff
@@ -497,7 +532,27 @@ def _castAttrValue(self, op, query, value):
...
 
+    def _invalidateCachedProperties(self):
+        """Invalidate all cached data property values."""
+        cached_props = getattr(self.__class__, '_cached_data_properties', set())
+
+        for prop_name in cached_props:
+            if prop_name in self.__dict__:
+                del self.__dict__[prop_name]
+
+    def _invalidateCacheAndLoadData(self, data):
```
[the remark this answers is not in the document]
JonnyWong16: `_loadData()` in `__init__` still needs to be changed?

pkkid:
I think the permissions were fixed.
image

JonnyWong16:
I found one more instance of `_loadData()` here. But I don't think this one matters because there are not cached `Setting` attributes?

https://github.com/pushingkarmaorg/python-plexapi/blob/65303c2a3b01be735dc1dd240231ba2f4bd40302/plexapi/settings.py#L40

> Also, this was another comment that was lost for the same reason:
> 
> `library.Hub` has a special `reload` function that calls `self.items = self.fetchItems(self.key)`, making the cached property somewhat useless if the `more and key` condition is passed.
> 
> I'm a little confused on why the remaining items are fetched only after a reload, why isn't it fetched in _loadData?
> 
> https://github.com/eliasbenb/python-plexapi/blob/88e165f5edb22b1d5b6528993b6989babfa2f64b/plexapi/library.py#L2236-L2248

If you do `LibrarySection.hubs()` it only returns "partial hubs" with just the first few items. You need to `reload()` the hub to retrieve the full list of items.

JonnyWong16:
> I'm assuming this was a requirement because getting the full list was too expensive before? If that's the case, would it make sense now that the attribute is lazy loaded to always return the full hub when called?

I think that's reasonable now.

JonnyWong16:
You're right, for some reason I thought `reload()` returned the items.

I'm working on another PR to refactor a bunch of the reload methods to unify them so just leave it for now.

pkkid (approved):

plexapi/base.py
```diff
@@ -945,21 +1001,42 @@ class PlexSession(object):
...
-        self.player = self.findItem(data, etag='Player')
-        self.session = self.findItem(data, etag='Session')
         self.sessionKey = utils.cast(int, data.attrib.get('sessionKey'))
-        self.transcodeSession = self.findItem(data, etag='TranscodeSession')
 
         user = data.find('User')
         self._username = user.attrib.get('title')
         self._userId = utils.cast(int, user.attrib.get('id'))
 
         # For backwards compatibility
```
glensc, line 1012: the "# For backwards compatibility" should have been retained in property methods

PR #1511, opened 2025-04-04
Defer XML string cleaning to improve performance

eliasbenb:
## Description

 This PR optimizes XML parsing by deferring the use of `utils.cleanXMLString`, which contains an expensive regex query. Instead of always cleaning the string before parsing, we now attempt to parse the XML string in its raw form, only falling back to cleaning if a `ParseError` occurs.

Originally suggested in https://github.com/pushingkarmaorg/python-plexapi/pull/1510#issuecomment-2777670823.

## Type of change

> None of the types of changes really apply, this is a performance enhancement

## Checklist:

### DIFF
--- a/plexapi/media.py
+++ b/plexapi/media.py
@@ -1075,7 +1075,7 @@ class BaseResource(PlexObject):
         data = f'{key}?url={quote_plus(self.ratingKey)}'
         try:
             self._server.query(data, method=self._server._session.put)
-        except xml.etree.ElementTree.ParseError:
+        except ElementTree.ParseError:
             pass
 
     @property


JonnyWong16:
Replace this with the helper function as well?

https://github.com/pushingkarmaorg/python-plexapi/blob/a8d051f29bbadca28dc9bd4bf50f1fb76d0fb77a/plexapi/myplex.py#L1883

plexapi/utils.py
```diff
@@ -718,3 +720,14 @@ def sha1hash(guid):
...
+
+def parseXMLString(s: str):
+    """ Parse an XML string and return an ElementTree object. """
+    if not s.strip():
+        return None
+    encoded_s = s.encode('utf-8')
+    try:  # Attempt to parse the string as-is without cleaning (which is expensive)
+        return ElementTree.fromstring(encoded_s)
+    except ElementTree.ParseError:  # If it fails, clean the string and try again
+        return ElementTree.fromstring(cleanXMLString(encoded_s))
```
JonnyWong16, line 733: Does `cleanXMLString` work on the encoded string? The original function cleaned then encoded.

plexapi/utils.py
```diff
@@ -23,6 +23,8 @@
 
 from plexapi.exceptions import BadRequest, NotFound, Unauthorized
 
+from xml.etree import ElementTree
```
JonnyWong16, line 26: Move standard library imports up.

JonnyWong16 (changes requested):
Could you also change `import xml` to `from xml.etree import ElementTree` in `media.py`.

https://github.com/pushingkarmaorg/python-plexapi/blob/a8d051f29bbadca28dc9bd4bf50f1fb76d0fb77a/plexapi/media.py#L2

And update this exception:

https://github.com/pushingkarmaorg/python-plexapi/blob/a8d051f29bbadca28dc9bd4bf50f1fb76d0fb77a/plexapi/media.py#L1078

plexapi/media.py
```diff
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
-import xml
 from pathlib import Path
 from urllib.parse import quote_plus
+from xml.etree import ElementTree
```
JonnyWong16, line 4: You need to update the exception below with the new import.

```py
        except ElementTree.ParseError: 
```

https://github.com/pushingkarmaorg/python-plexapi/blob/8894883cda07d2d72410f40fbde64e43937c69ed/plexapi/media.py#L1078

JonnyWong16 (approved):

PR #1576, opened 2026-01-16
Tear out Python 2 remnants

C0rn3j:
Remove all the `coding: utf-8` Python 2 kludges which were still present in most files, Python 3 does not need this.


JonnyWong16 (approved):
Thanks.

PR #1590, opened 2026-02-28
Datetime timezone awareness

eliasbenb:
## Description

Add optional timezone‑awareness for all parsed Plex datetime objects.

The PR introduces a `plexapi.timezone` config option (accepts `bool` or IANA time zone `str`). The configured time zone is parsed and globally defined. Subsequent calls to the `utils.toDatetime()` helper will normalize timestamps using that time zone.

Note: I've documented that "Toggling this option may break comparisons between aware and naive datetimes." Python's datetime does not support comparing naive and aware datetime objects (e.g. `some_tz_aware_dt >= some_naive_dt` will raise a `TypeError`).

## Type of change

## Checklist:

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -1,7 +1,9 @@
 import time
 
-import plexapi.utils as utils
 import pytest
+
+import plexapi
+import plexapi.utils as utils
 from plexapi.exceptions import NotFound
 
 
@@ -37,6 +39,21 @@ def test_utils_setDatetimeTimezone_local_and_invalid():
 
         assert utils.setDatetimeTimezone("Not/A_Real_Timezone") is None
         assert utils.toDatetime("0").tzinfo is None
+    finally:  # Restore for other tests
+        utils.DATETIME_TIMEZONE = original_tz
+
+
+def test_utils_package_datetime_timezone_stays_synced():
+    original_tz = utils.DATETIME_TIMEZONE
+    try:
+        tzinfo = utils.setDatetimeTimezone("UTC")
+        assert tzinfo is not None
+        assert plexapi.DATETIME_TIMEZONE is tzinfo
+
+        assert plexapi.DATETIME_TIMEZONE is utils.DATETIME_TIMEZONE
+        utils.setDatetimeTimezone(False)
+        assert plexapi.DATETIME_TIMEZONE is None
+        assert plexapi.DATETIME_TIMEZONE is utils.DATETIME_TIMEZONE
     finally:  # Restore for other tests
         utils.DATETIME_TIMEZONE = original_tz
 


plexapi/__init__.py
```diff
@@ -17,6 +17,7 @@
 PROJECT = 'PlexAPI'
 VERSION = __version__ = const.__version__
 TIMEOUT = CONFIG.get('plexapi.timeout', 30, int)
+DATETIME_TIMEZONE = setDatetimeTimezone(CONFIG.get('plexapi.timezone', False))
```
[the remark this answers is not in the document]
eliasbenb: This is incorrect. `CONFIG.get()` itself calls `utils.cast()` which handles bool-like conversion.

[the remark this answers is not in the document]
JonnyWong16: You need to specify type `bool` for `utils.cast()` to work, but that would break the use of `'local'` or IANA string. Bool-like values are handled inside `setDatetimeTimezone()`.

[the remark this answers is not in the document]
JonnyWong16: I agree with this, having `DATETIME_TIMEZONE` in two locations is confusing and prone to issues if the timezone is set again. 

If we want to keep the package-level access to `DATETIME_TIMEZONE`, just set the timezone at initialization time (this will set `utils.DATETIME_TIMEZONE`).

```suggestion
setDatetimeTimezone(CONFIG.get('plexapi.timezone', False))
```

Then add `__getattr__()` at the bottom of the file to alias `plexapi.DATETIME_TIMEZONE` to `plexapi.utils.DATETIME_TIMEZONE`.

```python
def __getattr__(name):
    """ Dynamic module attribute access for aliased values. """
    if name == 'DATETIME_TIMEZONE':
        return plexapi.utils.DATETIME_TIMEZONE
    raise AttributeError(f"module '{__name__}' has no attribute '{name}'")
```

Alternatively, we could just get rid of the package-level `DATETIME_TIMEZONE`, but it is convenient to have access to all your PlexAPI config values in the same place.

plexapi/utils.py
```diff
@@ -326,6 +330,67 @@ def threaded(callback, listargs):
...
+        if lower in ('true', '1'):
+            tzinfo = datetime.now().astimezone().tzinfo
+        elif lower in ('false', '0'):
+            tzinfo = None
+        else:
+            try:
+                tzinfo = ZoneInfo(setting)
+            except ZoneInfoNotFoundError:
+                tzinfo = None
+                log.warning('Failed to set timezone to "%s", defaulting to None', value)
```
JonnyWong16, line 366: Reuse `utils.cast()` for checking the bool config value.

```suggestion
    # Disable timezone if value is False or None
    if value is None or value is False:
        tzinfo = None
    # Use local timezone if value is True or "local"
    elif value is True or str(value).strip().lower() == 'local':
        tzinfo = datetime.now().astimezone().tzinfo
    # Attempt to resolve value as a boolean-like string or IANA timezone string
    else:
        setting = str(value).strip()
        # Try to cast as boolean first (normalize to lowercase for case-insensitive matching)
        try:
            is_enabled = cast(bool, setting.lower())
            tzinfo = datetime.now().astimezone().tzinfo if is_enabled else None
        except ValueError:
            # Not a boolean string, try parsing as IANA timezone
            try:
                tzinfo = ZoneInfo(setting)
            except ZoneInfoNotFoundError:
                tzinfo = None
                log.warning('Failed to set timezone to "%s", defaulting to None', value)
```

plexapi/utils.py
```diff
@@ -334,26 +392,16 @@ def toDatetime(value, format=None):
             format (str): Format to pass strftime (optional; if value is a str).
     """
     if value is not None:
+        tzinfo = DATETIME_TIMEZONE
         if format:
             try:
-                return datetime.strptime(value, format)
+                dt = datetime.strptime(value, format)
+                return dt.replace(tzinfo=tzinfo) if tzinfo else dt
             except ValueError:
```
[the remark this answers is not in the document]
JonnyWong16: `dt.tzinfo is not None` checks for UTC/offset already.

Could you add to your tests a format string containing a `%z` offset to confirm the behaviour is correct?

docs/configuration.rst
```diff
@@ -69,6 +70,19 @@ Section [plexapi] Options
...
+**timezone**
+    Controls whether :func:`~plexapi.utils.toDatetime` returns timezone-aware datetime objects.
+
+    * `false` (default): keep naive datetime objects (backward compatible).
+    * `true` or `local`: use the local machine timezone.
+    * IANA timezone string (for example `UTC` or `America/New_York`): use that timezone.
+
+    This feature relies on Python's :class:`zoneinfo.ZoneInfo` and the availability of IANA tzdata
+    on the system. On platforms without system tzdata (notably Windows), you may need to install
+    the :mod:`tzdata` Python package for IANA timezone strings (such as ``America/New_York``) to
```
JonnyWong16, line 82: Should we add `tzdata` as a requirement for Windows? Or just instruct the user to install it manually if they need it?

eliasbenb: The only real downside to adding tzdata is that it's an extra, probably unnecessary dependency.

However, besides Windows, I thought I should also note that Alpine Linux doesn't have tzdata installed system-wide by default either, and that's a pretty big user base, I imagine.

It's up to you, but I'd tack the requirement on.

You could also make it a conditional dependency like:

```toml
dependencies = [
  'tzdata; platform_system == "Windows"',
]
```

Although, I'm not aware of a way to do this for Alpine.

plexapi/__init__.py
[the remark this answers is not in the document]
eliasbenb: I decided to just re-export the `DATETIME_TIMEZONE` import rather than handle the attribute dynamically.


```py
from plexapi.utils import SecretsFilter, setDatetimeTimezone, DATETIME_TIMEZONE  # noqa: F401 # Re-export DATETIME_TIMEZONE
```

plexapi/__init__.py
```diff
@@ -6,7 +6,7 @@
 
 from plexapi.config import PlexConfig, reset_base_headers
 import plexapi.const as const
-from plexapi.utils import SecretsFilter
+from plexapi.utils import SecretsFilter, setDatetimeTimezone, DATETIME_TIMEZONE  # noqa: F401 # Re-export DATETIME_TIMEZONE
```
JonnyWong16: `DATETIME_TIMEZONE` is desynced from `utils.DATETIME_TIMEZONE` if `setDattimeTimezone()` is called again later.

eliasbenb: Sorry, you're right. You brought this up a little earlier but I was under the impression a re-export would point to the same DATETIME_TIMEZONE reference. Thinking through it now, strings are immutable in Python, so that wouldn't make sense.

I'll go ahead with your original suggestion of dynamic attribute loading.

docs/configuration.rst
JonnyWong16: I say we just leave it.

plexapi/__init__.py
```diff
@@ -50,3 +53,10 @@
 logfilter = SecretsFilter()
 if CONFIG.get('log.show_secrets', '').lower() != 'true':
     log.addFilter(logfilter)
+
+
+def __getattr__(name):
+    """ Dynamic module attribute access for aliased values. """
+    if name == 'DATETIME_TIMEZONE':
+        return utils.DATETIME_TIMEZONE
```
JonnyWong16, line 61: One last thing.

Add a test to check `plexapi.DATETIME_TIMEZONE` stays in sync with `utils.DATETIME_TIMEZONE`.

eliasbenb: Good call.

JonnyWong16 (approved):
Looks good. Thanks.

PR #1621, opened 2026-07-02
Allow query parameters in LibrarySection.hubs() and add Playlist.centroid

ajacobson:
## Description

Two small additions to support the personalized "Mixes For You" hub in music sections:

- `LibrarySection.hubs()` now accepts optional query parameters (e.g. `count=10` to limit items per hub, or `includeMyMixes=1` to include the "Mixes For You" hub), matching the signature `Library.hubs()` already has. `includeStations=1` remains the default via `setdefault`, so the request is unchanged for existing callers (including `MusicSection.stations()`), and it can now be overridden.

- New `Playlist.centroid` property returning the centroid `Artist` a personalized "Mix For You" playlist is built around, or `None` for regular playlists. Mix playlists have no `composite`/`thumb` of their own — the nested centroid `<Directory>` is their only artwork source, and there was previously no public way to reach it.

- `'centroid'` is added to `_DONT_RELOAD_FOR_KEYS`: it is legitimately `None` on every regular playlist, so without the exemption, accessing it on a partial object would trigger a spurious auto-reload (which can never surface a centroid anyway — it only exists in the hub payload). The mocked test asserts zero reloads.

Since the CI bootstrap server cannot generate personalized mixes, `test_Playlist_centroid` replays a hubs response captured from a live PMS (attribute shapes preserved: mixes carry no `ratingKey`, and artwork lives only on the nested centroid `<Directory>`).

Tested against a live PMS: default `hubs()` output unchanged across section types, `MusicSection.stations()` unaffected, `count`/include flags honored on movie/show/music sections, and 9 real mixes parsing with centroid artist + thumb with no extra requests.

Motivation: music-assistant/server#3736, where the downstream Plex provider currently maintains a manual-parsing workaround for these items and the maintainers asked for an upstream fix.

## Type of change

## Checklist:

### DIFF
--- a/plexapi/library.py
+++ b/plexapi/library.py
@@ -710,10 +710,17 @@ class LibrarySection(PlexObject):
         key = f'/hubs/sections/{self.key}/manage'
         self._server.query(key, method=self._server._session.delete)
 
-    def hubs(self):
+    def hubs(self, **kwargs):
         """ Returns a list of available :class:`~plexapi.library.Hub` for this library section.
+
+            Parameters:
+                **kwargs (dict): Optional query parameters to add to the request
+                    (e.g. ``count=10`` to limit the number of items per hub or
+                    ``includeMyMixes=1`` to include the personalized "Mixes For You" hub).
+                    ``includeStations=1`` is included by default and may be overridden.
         """
-        key = self._buildQueryKey(f'/hubs/sections/{self.key}', includeStations=1)
+        kwargs.setdefault('includeStations', 1)
+        key = self._buildQueryKey(f'/hubs/sections/{self.key}', **kwargs)
         return self.fetchItems(key)
 
     def agents(self):
--- a/plexapi/library.py
+++ b/plexapi/library.py
@@ -716,10 +716,12 @@ class LibrarySection(PlexObject):
             Parameters:
                 **kwargs (dict): Optional query parameters to add to the request
                     (e.g. ``count=10`` to limit the number of items per hub or
-                    ``includeMyMixes=1`` to include the personalized "Mixes For You" hub).
-                    ``includeStations=1`` is included by default and may be overridden.
+                    ``includeMyMixes=True`` to include the personalized "Mixes For You" hub).
+                    ``includeStations`` is included by default and may be overridden
+                    with ``includeStations=False``.
         """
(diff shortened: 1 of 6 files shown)


plexapi/library.py
```diff
@@ -710,10 +710,17 @@ def resetManagedHubs(self):
...
 
-    def hubs(self):
+    def hubs(self, **kwargs):
         """ Returns a list of available :class:`~plexapi.library.Hub` for this library section.
+
+            Parameters:
+                **kwargs (dict): Optional query parameters to add to the request
+                    (e.g. ``count=10`` to limit the number of items per hub or
+                    ``includeMyMixes=1`` to include the personalized "Mixes For You" hub).
+                    ``includeStations=1`` is included by default and may be overridden.
```
JonnyWong16: Other parts of the library allow passing in a `bool` for include parameters.

e.g. `includeMyMixes=True` or `includeStations=True`

JonnyWong16 (changes requested):

plexapi/library.py
ajacobson: Done — bools are now normalized to `1`/`0` (e.g. `includeMyMixes=True`, `includeStations=False`), and the tests exercise both. Docstring examples updated to match.

plexapi/library.py
```diff
@@ -710,10 +710,19 @@ def resetManagedHubs(self):
...
+
+            Parameters:
+                **kwargs (dict): Optional query parameters to add to the request
+                    (e.g. ``count=10`` to limit the number of items per hub or
+                    ``includeMyMixes=True`` to include the personalized "Mixes For You" hub).
+                    ``includeStations`` is included by default and may be overridden
+                    with ``includeStations=False``.
+        """
+        kwargs.setdefault('includeStations', 1)
+        kwargs = {k: 1 if v is True else 0 if v is False else v for k, v in kwargs.items()}
```
JonnyWong16: A bit cleaner


```suggestion
        kwargs = {k: int(v) if isinstance(v, bool) else v for k, v in kwargs.items()}
```

JonnyWong16 (changes requested):

JonnyWong16 (approved):
Looks good. Thanks.

